PR #11, opened 2019-07-15
add iso tests

### LINKED ISSUE #7: Test conversions for message types
We wrote all these wrappers for the protobuf generated files, and `Iso`s  for converting back and forth, but we also will need to write tests that go back and forth between these types to make sure they are indeed `Iso`s.

I gave `Generic` instances to all the message types and their fields so that we can easily derive quickcheck `Arbitrary` instances for these tests. But those could only be used to help prove that the composition `WrappedType -> ProtobufType -> WrappedType` is the `identity`. In other words, our Wrapped types embed into the protobuf types. In order to go the other way around we would need to give arbitrary instances to the protobuf types as well...

using quickcheck: 

- [x] test that the conversion    `WrappedType -> ProtobufType -> WrappedType` is identity
- [x] test that convervsion  `ProtobufType -> WrappedType -> ProtbufType`

safareli:
fix #7


martyall:
I got the tests passing for the fields. You'll want to use `isoCheck'` to finish with the complete message types, or maybe you can see something more clever.

safareli:
@blinky3713 unless you have any comments we can merge this

PR #16, opened 2019-07-17
Charlescrain/issue 8 correct protobufs

charlescrain:
## Changes

* updated `types.proto` file to use https://github.com/tendermint/tendermint/blob/v0.32.0/abci/types/types.proto
* Added new `data Event` in `FieldTypes.hs` 
* renamed `data BlockSizeParams` -> `data BlockParams`
* Added `isoCheck` tests for `Event`

## Note 

The previous version of `types.proto` file had

```proto
option (gogoproto.marshaler_all) = true;
option (gogoproto.unmarshaler_all) = true;
option (gogoproto.sizer_all) = true;
option (gogoproto.goproto_registration) = true;
```

commented out, wasn't sure if we wanted this or not.

### DIFF
--- a/protos/types.proto
+++ b/protos/types.proto
@@ -12,14 +12,14 @@ import "vendored/tendermint/tendermint/crypto/merkle/merkle.proto";
 // NOTE: When using custom types, mind the warnings.
 // https://github.com/gogo/protobuf/blob/master/custom_types.md#warnings-and-issues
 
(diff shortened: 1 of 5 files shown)


martyall:
with regard to the commenting out, I think I forgot to uncomment after playing with stuff trying to make it compile. It's best to go ahead and put it back in 😬

martyall (approved):
🎉

PR #17, opened 2019-07-19
copy conduit

### LINKED ISSUE #14: Copy Pasta server conduit setup from original hs-abci
- [x] Bring in `App m` type from hs-abci [link](https://github.com/f-o-a-m/hs-abci/blob/master/src/Network/ABCI/Types.hs#L82)
- [x] Bring in encoding / decoding conduits
- [ ] look for possible improvements

by improvements, look for places where decoding fails or prefix decoding fails, or App fails (since App runs in IO)

let's not make the improvements, just make issues / comments for them

https://github.com/f-o-a-m/hs-abci/blob/master/src/Network/ABCI.hs#L59

safareli:
fix #14


stack.yaml
```diff
@@ -38,6 +38,12 @@ packages:
 # using the same syntax as the packages field.
 # (e.g., acme-missiles-0.3)
 # extra-deps: []
+extra-deps:
```
safareli, line 41: Without this changes I was getting:
```
/hs-abci-server/src/Network/ABCI/Internal/Wire.hs:30:1: error:
    Could not load module ‘Data.ProtoLens.Encoding.Bytes’
    it is a hidden module in the package ‘proto-lens-0.4.0.1’
    Use -v to see a list of the files searched for.
   |
30 | import           Data.ProtoLens.Encoding.Bytes (getVarInt, putVarInt,
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^...
```

martyall (changes requested):
The copy paste code is fine, but you answer some of the questions about where things look like they could use improvement? For example in how we parse things, or where errors are handled /reported in a way which we could do better We want to turn those into issues.

safareli:
Regarding improvements, at this moment i don't see anything important, I guess once we start actually building some examples on top of it we would have better understanding and will see useful improvements.

martyall (approved):

PR #23, opened 2019-07-23
clean up withProto and add errors

### LINKED ISSUE #18: make typed errors for conduits
right now we throw into `String` or use `Nothing`

1. in Wire https://github.com/f-o-a-m/hs-abci-server/blob/master/src/Network/ABCI/Internal/Wire.hs#L31
2. here (implicitly) https://github.com/f-o-a-m/hs-abci-server/blob/master/src/Network/ABCI/Types/Messages/Request.hs#L60
3. what kind of errors are produced here https://github.com/f-o-a-m/hs-abci-server/blob/master/src/Network/ABCI.hs#L67 ?

We should throw into an error type with more information before [responding](https://github.com/f-o-a-m/hs-abci-server/blob/master/src/Network/ABCI.hs#L82)

safareli:
fix #18

### DIFF
--- a/src/Network/ABCI/Internal/Wire.hs
+++ b/src/Network/ABCI/Internal/Wire.hs
@@ -3,15 +3,15 @@ module Network.ABCI.Internal.Wire
   , decodeLengthPrefixC
   ) where
 
+import           Data.Bifunctor                (first)
 import qualified Data.ByteString               as BS
-import qualified Data.ByteString.Base16        as BS16
 import           Data.Conduit                  (ConduitT, awaitForever, yield)
 import qualified Data.Conduit.List             as CL
 import           Data.ProtoLens.Encoding.Bytes (getVarInt, putVarInt,
                                                 runBuilder, runParser,
                                                 signedInt64ToWord,
                                                 wordToSignedInt64)
-import           Data.String.Conversions       (cs)
+import qualified Network.ABCI.Types.Error      as Error
 
 -- | Transforms a stream of 'ByteString' to a stream of varlength-prefixed
 --   'ByteString's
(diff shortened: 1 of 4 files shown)


src/Network/ABCI/Types/Error.hs
```diff
@@ -0,0 +1,51 @@
+module Network.ABCI.Types.Error where
+
+import qualified Data.ByteString         as BS
+import qualified Data.ByteString.Base16  as BS16
+import qualified Data.ProtoLens          as PL
+import           Data.String.Conversions (cs)
+
+data Error
+  = CanNotDecodeRequest
+      { canNotDecodeRequestRequest :: BS.ByteString
```
safareli, line 10: it's ugly, let me know if we can do better here

src/Network/ABCI/Internal/Wire.hs
```diff
@@ -28,22 +28,19 @@ encodeLengthPrefixC = CL.map $ foldMap encodeLengthPrefix
...
+    splitOffMessages :: BS.ByteString -> Either Error.Error [BS.ByteString]
     splitOffMessages bs
-      | bs == mempty = pure []
+      | bs == mempty = Right []
       | otherwise = do
-          n <- runParser getVarInt bs
+          n <- first (Error.ProtoLensParseError bs) $ runParser getVarInt bs
           let lengthHeader = runBuilder $ putVarInt n
           messageBytesWithTail <- case BS.stripPrefix lengthHeader bs of
-            Nothing -> let prefixWithMsg = show (cs . BS16.encode $ lengthHeader :: String , cs . BS16.encode $ bs :: String)
```
safareli, line 44: Why was this BS16.encode used?
I  used it in showBS too if it's not needed LMK

martyall: It's used because if you print raw `ByteString` it comes out looking like nonsense, for readablity people always use an encoding, and to me `Hex` encoding is the best for humans

src/Network/ABCI/Types/Error.hs
martyall: seems fine to me

martyall (commented):
nice work 🇬🇪

martyall:
make sure you run stylish haskell and hlint before merging

PR #24, opened 2019-07-23
Natural transformation

### LINKED ISSUE #20: Make the conduit run in IO
Right now we have `type App m = ...`, and this `m` leaks everywhere even though we're never using it. Instead, we should require the `run` function to take a natural transformation `m ~> IO` and apply that in the one place we are running that application. Change all the types needed so that [this conduit](https://github.com/f-o-a-m/hs-abci-server/blob/master/src/Network/ABCI.hs#L63) runs in `IO` to prevent needless abstraction but also it will probably be faster.

### LINKED ISSUE #21: make main conduit run in IO
Right now we have `type App m = ...`, and this `m` leaks everywhere even though we're never using it. Instead, we should require the `run` function to take a natural transformation ` forall a. m a ~> IO a` and apply that in the one place we are running that application. Change all the types needed so that [this conduit](https://github.com/f-o-a-m/hs-abci-server/blob/master/src/Network/ABCI.hs#L63) runs in `IO` to prevent needless abstraction but also it will probably be faster.

martyall:
- fixes #21
- selective exports at top level
- add modules to `expose-modules`
- remove dependency on `network` by adding `onAquire :: AppData -> IO ()` argument
- get ride of unused dependencies

### DIFF
--- a/src/Network/ABCI.hs
+++ b/src/Network/ABCI.hs
@@ -1,11 +1,14 @@
-module Network.ABCI where
+module Network.ABCI
+  ( defaultLocalSettings
+  , serveAppWith
+  , serveApp
+  ) where
 
 import           Data.Conduit                         (ConduitT, runConduit,
                                                        (.|))
 import qualified Data.Conduit.List                    as CL
 import           Data.Conduit.Network                 (AppData, ServerSettings,
-                                                       appSink,
-                                                       appSource,
+                                                       appSink, appSource,
                                                        runGeneralTCPServer,
                                                        serverSettings)
 import           Data.Monoid                          ((<>))
--- a/src/Network/ABCI.hs
+++ b/src/Network/ABCI.hs
@@ -24,11 +24,13 @@ import qualified Network.ABCI.Types.Messages.Response as Response
 import qualified Proto.Types                          as PT
 
 
--- | Default ABCI app network settings.
+-- | Default ABCI app network settings for serving on localhost at the
+-- standard port.
 defaultLocalSettings :: ServerSettings
 defaultLocalSettings = serverSettings 26658 $ fromString "127.0.0.1"
 
--- | Serve an ABCI application with custom 'ServerSettings'
+-- | Serve an ABCI application with custom 'ServerSettings' and a custom
+-- action to perform on aquiring the socket resource.
 serveAppWith
   :: ServerSettings
   -> App IO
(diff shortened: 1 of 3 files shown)


charlescrain (approved):
LGTM

src/Network/ABCI/Types/App.hs
```diff
@@ -1,8 +1,21 @@
...
 
-newtype App m =
-  App (forall (t :: MessageType). Request t -> m (Response t))
+-- | Application type that represents a well typed application, i.e. a
+-- function from a typed `Request` to a typed `Response`.
+newtype App m = App
+  { runApp :: forall (t :: MessageType). Request t -> m (Response t) }
+
+-- | Transform an application from running in a custom monad to running in `IO`.
+transformApp
```
safareli, line 16: type of this function can be more general, why specialize to IO?
```
transformApp
  :: (forall a. m a -> g a)
  -> App m
  -> App g
```

src/Network/ABCI.hs
```diff
@@ -20,47 +21,37 @@ import           Network.ABCI.Internal.Wire           as Wire
...
-     )
-  => ServerSettings
-  -> (SockAddr -> m (App m))
-  -> m ()
-serveAppWith cfg mkApp = runGeneralTCPServer cfg $ \appData -> do
-  app <- mkApp (appSockAddr appData)
-  runConduit (setupConduit app appData)
+  :: ServerSettings
+  -> App IO
+  -> (AppData -> IO ())
```
safareli, line 37: I suspect when AppData is useful, you would need it in the App itself, and with this signature, it would be hard to obtain it in the App (you would need to use some MVar to put AppData into, and then read from it in the App). 
Is there any disadvantage for `AppData -> IO (App IO)`?

safareli:
I suspect it also fixes #21

src/Network/ABCI.hs
martyall: It might be too early to tell, but I somehow doubt that this `AppData` would be useful at the level of `App` creation. You can find its [definition](https://hackage.haskell.org/package/streaming-commons-0.2.1.0/docs/src/Data.Streaming.Network.Internal.html#AppData) in the [streaming-commons](https://hackage.haskell.org/package/streaming-commons-0.2.1.0) lib. If the `App` is using this, it means it's reading / writing from the socket directly without the preprocessing conduits, or possibly close the connection itself. I'm not sure we need any of these features, but who knows. Delay decision till later? 

The reason I even wanted to expose it at this level is for testing, since I think we can write messages to the socket directly with [this](https://hackage.haskell.org/package/streaming-commons-0.2.1.0/docs/Data-Streaming-Network.html#v:writeLens). And yeah we can share the AppData with the test suite via an `MVar`.

src/Network/ABCI/Types/App.hs
martyall: good point

src/Network/ABCI.hs
safareli: good point take a look at  #26

safareli (approved):

PR #25, opened 2019-07-24
add tests

### LINKED ISSUE #15: Write tests for conduit-backed server
For unit tests, look at the old tests from https://github.com/f-o-a-m/hs-abci/blob/master/test/Network/ABCI/Internal/WireSpec.hs and see which ones we actually care about. 

set up some basic tests which actually uses a server? What should this do?

safareli:
fix #15


safareli:
regarding
> set up some basic tests which actually uses a server? What should this do?

Because we are hard coding IO into the coduint (#21) It would be harder to test it

safareli:
#28

PR #26, opened 2019-07-24
thoughts?

safareli:
I started this off the err branch, and target is that same branch to make diff nicer:

This snippet demonstrates how things play together:

```hs
app :: App (Free DSL)
interpret :: Config -> Free DSL ~> IO
mkConfig :: IO Config

main = do
  conf <- mkConfig appData
  serveApp $ transformApp (interpret conf) app
```

### DIFF
--- a/src/Network/ABCI.hs
+++ b/src/Network/ABCI.hs
@@ -7,7 +7,7 @@ import           Control.Monad.Trans.Except           (runExceptT, except)
 import qualified Data.ByteString                      as BS
 import           Data.Conduit                         (runConduit, (.|))
 import qualified Data.Conduit.List                    as CL
-import           Data.Conduit.Network                 (AppData, ServerSettings,
+import           Data.Conduit.Network                 (ServerSettings,
                                                        appSink,
                                                        appSource,
                                                        runTCPServer,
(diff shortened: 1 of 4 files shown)


src/Network/ABCI.hs
```diff
@@ -21,74 +22,47 @@ import           Network.ABCI.Types.App               (App (..))
...
-  app <- mkApp (appSockAddr appData)
-  runConduit (setupConduit app appData)
+  :: ServerSettings
+  -> (AppData -> IO (AppRunner IO))
+  -> IO ()
+serveAppWith cfg mkRunner = runTCPServer cfg $ \appData -> do
+  runner <- mkRunner appData
+  runConduit $
+      appSource appData
+    .| CL.mapM runner
```
safareli, line 39: We just use one mapM now as all other transformations were bunch of maps and one mapM and they can be fused together.

src/Network/ABCI.hs
```diff
@@ -21,74 +22,47 @@ import           Network.ABCI.Types.App               (App (..))
...
-     appSource appData
-  .| Wire.decodeLengthPrefixC
-  .| CL.map (traverse decodeRequestValue =<<)
-  .| CL.mapM (respondWith app)
-  .| CL.map (map PL.encodeMessage)
-  .| Wire.encodeLengthPrefixC
-  .| appSink appData
+runApp :: Monad m => App m -> AppRunner m
+runApp (App app) bsInput = do
+  responcesE <- runExceptT $ do
```
safareli, line 49: ExceptT is used so we don have functions like `Either Error x -> ....`

src/Network/ABCI.hs
```diff
@@ -21,74 +22,47 @@ import           Network.ABCI.Types.App               (App (..))
...
-  -> ConduitT i o m ()
-setupConduit app appData =
-     appSource appData
-  .| Wire.decodeLengthPrefixC
-  .| CL.map (traverse decodeRequestValue =<<)
-  .| CL.mapM (respondWith app)
-  .| CL.map (map PL.encodeMessage)
-  .| Wire.encodeLengthPrefixC
-  .| appSink appData
+runApp :: Monad m => App m -> AppRunner m
```
safareli, line 47: it's pure function and can make testing easier

safareli:
This is good point https://github.com/f-o-a-m/hs-abci-server/pull/24#discussion_r306934838 so removed the AppData all together as for testing one can just use `runApp`

src/Network/ABCI.hs
```diff
@@ -21,74 +22,41 @@ import           Network.ABCI.Types.App               (App (..))
...
```
safareli, line 47: With current arrangement I would say to even move runApp and decodeLengthPrefix/encodeLengthPrefix to App module

martyall:
(thoughts given on rocket chat)

safareli:
#28

PR #28, opened 2019-07-25
Use less coduint and add tests

### LINKED ISSUE #15: Write tests for conduit-backed server

safareli:
fix #15


src/Network/ABCI/Types/App.hs
```diff
@@ -1,21 +1,89 @@
...
+    Nothing -> Left $ DecodeError.NoValueInRequest packet (request ^. PL.unknownFields)
+    Just value -> Right $ value
+
+-- | Encodes ByteStrings into varlength-prefixed ByteString
+encodeLengthPrefix :: [BS.ByteString] -> LengthPrefixedByteString
+encodeLengthPrefix = foldMap $ \bytes ->
+  let headerN = signedInt64ToWord . fromIntegral . BS.length $ bytes
+      header = runBuilder $ putVarInt headerN
+  in header `BS.append` bytes
+{-# INLINEABLE encodeLengthPrefix #-}
```
safareli: Do we still need this INLINEABLE ? if yes shuold we add it to other functions too?

src/Network/ABCI/Types/App.hs
```diff
@@ -1,21 +1,89 @@
...
+runApp (App app) bs =
+  bs
+    & (decodeLengthPrefix >=> decodeRequests)
+    & either (pure . onError) (traverse onResponce)
+    & fmap (encodeLengthPrefix . encodeResponces)
+  where
+    onError :: DecodeError -> [PT.Response]
+    onError err = [Response.toProto $ Response.ResponseException $ Response.Exception $ cs $ DecodeError.print err]
+    -- for some reason with this type we get • Couldn't match type ‘m1’ with ‘m’
+    -- onResponce :: PT.Request'Value -> m PT.Response
```
safareli: Any ideas on this?

test/Network/ABCI/Test/Types/AppSpec.hs
```diff
@@ -0,0 +1,80 @@
...
+    -- | This property test (see https://en.wikipedia.org/wiki/QuickCheck)
+    -- for pointers on what this is) checks that these functions are the
+    -- inverse of each other for arbitrarily chunked arbitrary 'ByteString's.
+    -- When the 'Producer' is a real socket, each chunk is the result of a
+    -- 'recv' call. We check for robustness against a peer that decides to
+    -- flush their side of the connection in funky ways.
+
+    it "decoding an encoded bytestring yields the same bytestring" $
+      property $ \(bytelist, nonNegativeChunkSizes) ->
+        let conduit = chunksProducer bytes nonNegativeChunkSizes
```
safareli: I suspect there is no need to use `nonNegativeChunkSizes` as in we can just generate arbitrary [ByteString] and size of each bytestring will be random anyway. Am I missing something?

src/Network/ABCI/Types/App.hs
```diff
@@ -1,21 +1,89 @@
...
-  nat $ f req
+transformApp :: (forall a. m a -> g a) -> App m -> App g
+transformApp nat (App f) = App $ nat . f
+
+-- | compiles `App` down to `AppBS`
+runApp :: Applicative m => App m -> AppBS m
+runApp (App app) bs =
+  bs
+    & (decodeLengthPrefix >=> decodeRequests)
+    & either (pure . onError) (traverse onResponce)
```
martyall: spelling error: Resonce -> Response

there are other places where this appears

src/Network/ABCI/Types/App.hs
martyall: I don't think we need any of these pragmas until we actually take a look at a core dump. It's probably a good exercise in ghc-ism. Let's go ahead and remove them and do a complete pass of this when we build the example app.

src/Network/ABCI/Types/App.hs
```diff
@@ -1,21 +1,89 @@
...
+    & fmap (encodeLengthPrefix . encodeResponces)
+  where
+    onError :: DecodeError -> [PT.Response]
+    onError err = [Response.toProto $ Response.ResponseException $ Response.Exception $ cs $ DecodeError.print err]
+    -- for some reason with this type we get • Couldn't match type ‘m1’ with ‘m’
+    -- onResponce :: PT.Request'Value -> m PT.Response
+    onResponce = Request.withProto $ fmap Response.toProto . app
+
+-- | ByteString which contains multiple length prefixed messages
+type LengthPrefixedByteString = BS.ByteString
```
martyall: I think a type alias in this case doesn't add a lot of value, I would say either newtype it or keep it as `ByteString`. But since we don't have to parse anything into this type, maybe it's just a raw `ByteString`?

 In the case of newtype, don't expose the constructor.

src/Network/ABCI/Types/App.hs
```diff
@@ -1,21 +1,89 @@
...
+
+-- | compiled Application which operates on varlength-prefixed ByteString
+type AppBS m = LengthPrefixedByteString -> m LengthPrefixedByteString
+
+
+-- | Encodes responces to bytestrings
+encodeResponces :: [PT.Response] -> [BS.ByteString]
```
martyall: This is only used in one other function and is relatively simple, consider including it in a `where` clause there.

src/Network/ABCI/Types/App.hs
```diff
@@ -1,21 +1,89 @@
...
+    onError err = [Response.toProto $ Response.ResponseException $ Response.Exception $ cs $ DecodeError.print err]
+    -- for some reason with this type we get • Couldn't match type ‘m1’ with ‘m’
+    -- onResponce :: PT.Request'Value -> m PT.Response
+    onResponce = Request.withProto $ fmap Response.toProto . app
+
+-- | ByteString which contains multiple length prefixed messages
+type LengthPrefixedByteString = BS.ByteString
+
+-- | compiled Application which operates on varlength-prefixed ByteString
+type AppBS m = LengthPrefixedByteString -> m LengthPrefixedByteString
```
martyall: Is there a way to not expose this?

test/Network/ABCI/Test/Types/AppSpec.hs
```diff
@@ -0,0 +1,80 @@
...
+
+
+main :: IO ()
+main = hspec spec
+
+spec :: Spec
+spec = do
+  describe "decodeLengthPrefix / encodeLengthPrefix" $ do
+
+    -- | This property test (see https://en.wikipedia.org/wiki/QuickCheck)
```
martyall: remove this comment, it's not really relevant for the unit test

test/Network/ABCI/Test/Types/AppSpec.hs
```diff
@@ -0,0 +1,80 @@
...
+        in runConduitPure conduit == Right bytes
+
+  describe "decodeLengthPrefix" $ do
+
+    it "fails gracefully when given a string larger than maxMessageLen" $
+      let ginormousSizeVarLen = 8 `BS.cons` runPut (Put.putWord64be maxBound)
+          conduit = yield ginormousSizeVarLen
+                 .| CL.map (first DecodeError.print . decodeLengthPrefix)
+                 .| consumeValidChunks
+      in runConduitPure conduit `shouldSatisfy` isLeft
```
martyall: why do you need `first DecodeError.print` if you're just checking `isLeft` ?

test/Network/ABCI/Test/Types/AppSpec.hs
martyall: yes i agree, `chunksProducer` seems useless.

test/Network/ABCI/Test/Types/AppSpec.hs
```diff
@@ -0,0 +1,80 @@
...
+import           Data.Conduit
+import qualified Data.Conduit.List as CL
+import           Data.Either (isLeft)
+import           Data.Monoid ((<>))
+import           Test.Hspec
+import           Test.QuickCheck
+
+
+main :: IO ()
+main = hspec spec
```
martyall: this is not needed with the `hspec-discover pragma`

martyall (changes requested):

src/Network/ABCI/Types/App.hs
martyall: You could try passing `App m` in an argument to `onResponse` in the where clause and pattern match there. 

```haskell

onResponce :: App m -> PT.Request'Value -> m PT.Response


```

you might rename it to `buildResponse` or something, idk

src/Network/ABCI/Types/App.hs
safareli: removed

src/Network/ABCI/Types/App.hs
safareli: Adding `forall m.` to the definition of `runApp` fixed the issue

src/Network/ABCI/Types/App.hs
safareli: Created newtype but I had to export constructor

src/Network/ABCI/Types/App.hs
safareli: Yes, I've removed and inlined this

test/Network/ABCI/Test/Types/AppSpec.hs
safareli: DecodeError has no Show instance

martyall (approved):

Issue #22, opened 2019-07-23
example app


martyall:
We need to make a realistic example app, I'm thinking simple storage. Maybe we add the additional requirement that the update count messages contain an integer strictly greater than the current count. It should

- [x] use the logger middleware when it's available
- [x] have a docker file
- [ ] have e2e tests (possibly in PS) that use the docker network
- [ ] have benchmarks
- [x] use a `.proto` file for the messages

safareli:
partially fixed with  #34

martyall:
closing in favor of #40

Issue #19, opened 2019-07-23
Define middleware type and make a basic logging middleware


martyall:
we mentioned defining a middleware type as  `App m -> App m`

We're about to start having to write tests and a small example app, so we should go ahead and put in a basic logger middleware. Something that allows for location, severity, and some loggable fields. It should take `LoggingOptions` that control the severity / level of info / color?

martyall:
I assigned this to @charlescrain because he was talking about making use of `katip` instead of rolling our own whatever, which seems like the better option for getting something up and running.

however, I think this needs to be used in it's own cabal project inside this repo, something like `hs-abci-katip-logger`  or something like this, since people might not want to depend on the `katip` library. It would depend on `hs-abci-server` for the middleware type alias

```haskell
type Middleware = App IO -> App IO
```

which isn't defined yet, as well as the message types I guess.

charlescrain:
Great point, @blinky3713 the `wai` example you gave me for middleware also has the `wai-extra` library that has a bunch for different middleware available. We could do a similar thing and it could be a place for anyone to add useful middleware modules.

kejace:
FWIW I found this that seems like a good logging library: http://hackage.haskell.org/package/co-log

martyall:
katip looks pretty sweet cause it can also push to datadog, which we **LOVE**

charlescrain:
It's probably faster for me to use Katip since I'm more familiar with it but I can do whichever.

kejace:
My apologies I thought this issue was about middle-_management_ not middleware. What am I even doing here.

charlescrain:
@blinky3713 I've built out some basic scaffolding for the logger. I was planning on making the logger for Requests and modeling it after [wai-request-logger](http://hackage.haskell.org/package/wai-extra-3.0.27/docs/Network-Wai-Middleware-RequestLogger.html). 

I began making the instances to log the requests but wanted to ask if you guys are okay with JSON and other necessary orphan instances or if you'd prefer making newtype wrappers. 

Here's the [branch](https://github.com/f-o-a-m/hs-abci-server/tree/charles/logging-middleware) I've started.

martyall:
Looks good to me. I think that if we can get around orphans that's best, newtyping sounds good to me.

Also I think we should go ahead and nest `hs-abci-server` into a directory as well, so that only the stack yaml, license, and various READMEs Makefiles etc are top level. What do you think?

charlescrain:
Sounds good!

[...]

PR #29, opened 2019-07-26
Nest directories

martyall:
This is a PR for a filestystem restructure thats necessary for both #22 and #19


charlescrain (approved):
boom!

charlescrain:
This is great, didn't realize I'd need to add my sub dir to the stylish haskell make command :D

PR #32, opened 2019-07-29
Charles/logging middleware

charlescrain:


charlescrain:
I'm not sure why it wasn't successful. The travis `Make` command is failing but if it's already in master why would it fail now?

charlescrain:
Hmm last build failed due to connection to github. I don't seem to be able to re-run the build. Pretty confident it will pass though. Let me know what you guys think about the interface.

hs-abci-server/src/Network/ABCI/Types/Messages/Common.hs
```diff
@@ -0,0 +1,9 @@
+module Network.ABCI.Types.Messages.Common
+  ( defaultABCIOptions
+  ) where
+
+import           Data.Aeson        (Options)
+import           Data.Aeson.Casing (aesonDrop, camelCase)
+
+defaultABCIOptions :: String -> Options
+defaultABCIOptions prefix = aesonDrop (length prefix) camelCase
```
safareli: Let's instead use `snakeCase`, as i'm using this json instance for tendermint sdk bindings and in go people use snakeCase

hs-abci-extra/src/Network/ABCI/Middleware/RequestLogger.hs
```diff
@@ -1,23 +1,66 @@
...
+
+---------------------------------------------------------------------------
+-- | mkRequestLogger
+---------------------------------------------------------------------------
+-- | Request logger middleware for ABCI requests with custom Katip LogEnv
+-- and Namespace.
+mkRequestLogger :: (MonadIO m) => LogEnv -> Namespace -> Middleware m
+mkRequestLogger le ns (App app) = App $ \ req -> do
+  runKatipContextT le () ns $ logRequest req
+  app req
```
safareli, line 82: shouldn't we also log response?

hs-abci-extra/LICENSE
```diff
@@ -0,0 +1,30 @@
+Copyright Charles Crain (c) 2019
```
safareli: Should we have licenses in subfolders? 
Also as I remember we use Apache

martyall: We can remove this auto generated license since we have the one a level up

hs-abci-extra/src/Network/ABCI/Middleware/RequestLogger.hs
charlescrain: Yeah we could, I was using [wai-request-logger](http://hackage.haskell.org/package/wai-extra-3.0.27/docs/Network-Wai-Middleware-RequestLogger.html) as an example of middleware. I think the response logger would be it's own separate piece of middleware. Let me know if you guys think i should make that as well.

hs-abci-extra/LICENSE
charlescrain: Ahh yeah this was from `stack new`

charlescrain: Ahh yeah this was from `stack new`

safareli:
https://travis-ci.com/f-o-a-m/hs-abci/builds/121693863#L265
you might run do haskell stylish

hs-abci-extra/src/Network/ABCI/Middleware/RequestLogger.hs
```diff
@@ -0,0 +1,66 @@
...
+
+import           Network.ABCI.Types.App              (App (..), Middleware)
+import           Network.ABCI.Types.Messages.Request (Request)
+
+---------------------------------------------------------------------------
+-- | Types
+---------------------------------------------------------------------------
+-- | Loggable newtype wrapper
+newtype Loggable a = Loggable a
+instance ToJSON (Loggable (Request t)) where
```
martyall: IMO we don't actually need this instance as it's only used for the `ToObject` instance. You could instead vendor the default instance like 

```haskell
instance ToObject (Loggable (Request t)) where
    toObject (Loggable v) = case toJSON v of
      A.Object o -> o
      -- unreachable case
      _          -> mempty
```

hs-abci-extra/src/Network/ABCI/Middleware/RequestLogger.hs
```diff
@@ -0,0 +1,66 @@
...
+  toJSON (Loggable a) = toJSON a
+
+instance ToObject (Loggable (Request t))
+instance LogItem (Loggable (Request t)) where
```
martyall: I think we're going to want to have at least two levels of verbosity here. In my mind you would use the lower for the console logging scribe, and the higher for ES or whatever. This might require some manual instance (?) but I think it's worth it to not blow up the console with info. what do you think?

hs-abci-extra/test/Spec.hs
```diff
@@ -0,0 +1,2 @@
+main :: IO ()
```
martyall: we probably don't need tests here, so we can delete this file

hs-abci-server/src/Network/ABCI/Types/Messages/Request.hs
```diff
@@ -56,6 +62,22 @@ data Request (m :: MessageType) :: * where
...
+  toJSON (RequestSetOption v)  = toJSON v
+  toJSON (RequestInitChain v)  = toJSON v
+  toJSON (RequestQuery v)      = toJSON v
+  toJSON (RequestBeginBlock v) = toJSON v
+  toJSON (RequestCheckTx v)    = toJSON v
+  toJSON (RequestDeliverTx v)  = toJSON v
+  toJSON (RequestEndBlock v)   = toJSON v
+  toJSON (RequestCommit v)     = toJSON v
+
+instance FromJSON (Request (t :: MessageType)) where
```
martyall: we can remove this as there is currently no intended use for these instances. In general you can't write instances lke this anyway, you would need to write them individually like

```haskell
instance FromJSON (Request 'MTEcho) where
...
```

hs-tendermint/hs-tendermint.cabal
```diff
@@ -0,0 +1,49 @@
+cabal-version: 1.12
```
martyall: how did this get in here? I think you must have pulled in iraklis work?

martyall (changes requested):

hs-abci-extra/src/Network/ABCI/Middleware/RequestLogger.hs
charlescrain: Yeah sure, the verbosity changes what keys gets logged for each object. Is there a method for deciding what keys to the requests are less important than others?

hs-tendermint/hs-tendermint.cabal
charlescrain: Yeah not sure how I got this. I'll cherry pick out the commits

hs-abci-extra/src/Network/ABCI/Middleware/RequestLogger.hs
safareli: good point, and we can add it later.

safareli (approved):

martyall (approved):

PR #36, opened 2019-08-02
Tendermint

### LINKED ISSUE #31: write client library for tendermint core
https://tendermint.com/rpc

safareli:
fix #31

### DIFF
--- a/hs-tendermint/src/Network/Tendermint.hs
+++ b/hs-tendermint/src/Network/Tendermint.hs
@@ -1,0 +1,271 @@
+module Network.Tendermint where
+
+import           Control.Monad.Reader                   (ReaderT, runReaderT)
+import           Data.Aeson                             (FromJSON (..),
+                                                         ToJSON (..),
+                                                         genericParseJSON,
+                                                         genericToJSON)
+import qualified Data.Aeson                             as Aeson
+import           Data.Aeson.Casing                      (aesonDrop, pascalCase,
+                                                         snakeCase)
+import           Data.Int                               (Int64)
+import           Data.Text                              (Text)
+import           GHC.Generics                           (Generic)
+import qualified Network.HTTP.Simple                    as HTTP
+import qualified Network.Tendermint.Internal.RPCClient  as RPC
+-- import qualified Network.ABCI.Types.Messages.Request as Request
+import           Data.ByteArray.HexString               (HexString)
+import           Data.Word                              (Word32)
+import qualified Network.ABCI.Types.Messages.FieldTypes as FieldTypes
+import qualified Network.ABCI.Types.Messages.Response   as Response
+
+type TendermintM a = ReaderT RPC.Config IO a
+
+runTendermintM :: RPC.Config -> TendermintM a -> IO a
+runTendermintM = flip runReaderT
+
+defaultConfig :: RPC.Config
+defaultConfig = RPC.Config
+  $ HTTP.setRequestHost "localhost"
+  $ HTTP.setRequestPort 26657
+  $ HTTP.defaultRequest
+
+
+-- https://github.com/tendermint/tendermint/blob/14fa8007738c51fe0951544a63eeea759679d495/rpc/core/abci.go
+abciInfo :: TendermintM ResultABCIInfo
+abciInfo = RPC.remote (RPC.MethodName "abci_info") ()
+
+-- https://github.com/tendermint/tendermint/blob/14fa8007738c51fe0951544a63eeea759679d495/rpc/core/health.go#L35
+health :: TendermintM ResultHealth
+health = RPC.remote (RPC.MethodName "health") ()
+
+-- https://github.com/tendermint/tendermint/blob/14fa8007738c51fe0951544a63eeea759679d495/rpc/core/abci.go#L56
+abciQuery :: RequestABCIQuery -> TendermintM ResultABCIQuery
+abciQuery = RPC.remote (RPC.MethodName "abci_query")
+
+-- https://github.com/tendermint/tendermint/blob/14fa8007738c51fe0951544a63eeea759679d495/rpc/core/blocks.go#L72
+block :: RequestBlock -> TendermintM ResultBlock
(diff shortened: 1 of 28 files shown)


hs-tendermint/src/Network/Tendermint.hs
```diff
@@ -0,0 +1,271 @@
...
+broadcastTxSync = RPC.remote (RPC.MethodName "broadcast_tx_sync")
+
+-- https://github.com/tendermint/tendermint/blob/14fa8007738c51fe0951544a63eeea759679d495/rpc/core/mempool.go#L215
+broadcastTxCommit :: RequestBroadcastTxCommit -> TendermintM ResultBroadcastTxCommit
+broadcastTxCommit = RPC.remote (RPC.MethodName "broadcast_tx_commit")
+
+
+
+defaultRPCOptions :: String -> Aeson.Options
+defaultRPCOptions prefix = aesonDrop (length prefix) snakeCase
```
safareli: TODO: use one from Types.Common

hs-tendermint/src/Network/Tendermint/Internal/RPCClient.hs
```diff
@@ -0,0 +1,114 @@
...
+  , ToJSON input
+  )
+  => MethodName
+  -> input
+  -> m output
+{-# INLINE remote #-}
+remote method input = do
+  rid <- abs <$> liftIO randomIO
+  let req = Request method rid (toJSON input)
+  (Config baseHTTPRequest) <- ask
```
martyall: parens not needed

hs-tendermint/src/Network/Tendermint/Internal/RPCClient.hs
```diff
@@ -0,0 +1,114 @@
...
+data JsonRpcException
+  = ParsingException String
+  | CallException RpcError
+  deriving (Eq, Show)
+
+instance Exception JsonRpcException
+
+
+-- | Name of called method.
+newtype MethodName = MethodName Text deriving (Eq, Show)
```
martyall: if you derive `IsString` then you don't have to use the constructor in the RPC calls,

e.g. 

```
tx :: RequestTx -> TendermintM ResultTx
tx = RPC.remote "tx"
```

hs-tendermint/src/Network/Tendermint/Internal/RPCClient.hs
```diff
@@ -0,0 +1,114 @@
...
+  )
+  => MethodName
+  -> input
+  -> m output
+{-# INLINE remote #-}
+remote method input = do
+  rid <- abs <$> liftIO randomIO
+  let req = Request method rid (toJSON input)
+  (Config baseHTTPRequest) <- ask
+  responce <- liftIO
```
martyall: mispelling, `response`

stack.yaml
```diff
@@ -1,8 +1,10 @@
+# https://www.stackage.org/lts-13.27
 resolver: lts-13.27
 
 packages:
 - ./hs-abci-server
 - ./hs-abci-extra
+- ./hs-tendermint
```
martyall: I think we should change the package name to `hs-tendermint-client` and use the `Client`  path in all the modules

hs-tendermint/src/Network/Tendermint.hs
```diff
@@ -0,0 +1,271 @@
...
+  , voteTimestamp        :: FieldTypes.Timestamp
+  , voteValidatorAddress :: HexString
+  , voteValidatorIndex   :: Int
+  , voteSignature        :: HexString
+  } deriving (Eq, Show, Generic)
+instance FromJSON Vote where
+  parseJSON = genericParseJSON $ defaultRPCOptions "vote"
+
+-- https://github.com/tendermint/tendermint/blob/14fa8007738c51fe0951544a63eeea759679d495/types/tx.go#L19
+type Tx = HexString
```
martyall: In general this kind of newtyping doesn't really add very much value, I think it's fine to leave it as `HexString` since that's become synonymous with `ByteString`

martyall (changes requested):
Just a few small comments. Also if you're going to rely on `hs-abci-server` then we might consider splitting out a separate message types package. thoughts?

hs-tendermint/src/Network/Tendermint/Internal/RPCClient.hs
safareli: if we add IsString then is there any advantage over just using `type MethodName = Text`?
I added the newtype so by looking at the type of remote it's clear what you need to pass.

hs-tendermint/src/Network/Tendermint.hs
safareli: It's not a newtype, it's type alias. I tried to sort of port types from go 1 to 1. I think it would make it easy to debug stuff once things go wrong. do you think otherwise?

hs-tendermint/src/Network/Tendermint.hs
safareli: actually Types.Common is hidden and will leave defaultRPCOptions as is now.

hs-tendermint/src/Network/Tendermint.hs
martyall: sorry I meant aliasing . In general I think type aliasing does well when you’re able to supply type some type variables with defaults (like halogen does this I’m pretty sure), but usually in this case I’m not for it because you don’t gain type safety and there will be places where we refer to Txs by just HexString, like in the other packages. You can get pretty far with just using haddock , see the bottom of this page https://www.haskell.org/haddock/doc/html/ch03s02.html

hs-tendermint/src/Network/Tendermint/Internal/RPCClient.hs
martyall: Yeah I’m this case there might be nothing to gain since probably every caller will use a string literal to name the method

martyall (approved):

PR #54, opened 2019-08-08
updated readme and RequestLogger haddocks

charlescrain:
I used `wai-extra`'s readme as an example; it points to an example that uses the middleware so I did the same.

Appreciate feedback on the haddocks.


safareli (approved):

safareli:
we have strange error 
> The command "eval git clone --depth=50 --branch=hs-abci-extra-docs https://github.com/f-o-a-m/hs-abci.git f-o-a-m/hs-abci " failed 3 times.

charlescrain:
I think its just a network error on the machine.

martyall (approved):

PR #59, opened 2019-08-08
Example app docker

### LINKED ISSUE #41: make docker network for example app
We want to make a docker network for the example app, because as things evolve we will be using this more and more as a sandbox to test ideas, and the app will get more complicated.

For example, we want to make a datadog middleware, we need a stats server. Or we want the whole elk stack.

The docker network should build the app from source, every other image should be hosted somewhere else.

martyall:
fixes #41 

- adds a `Dockerfile` to hs-abci-example
- adds a `docker-compose.yaml` file to hs-abci-example

NOTE:
The Dockerfile builds the binaries in docker, but I was talking with @charlescrain and we thought it would be worth it to make a separate Dockerfile for building on linux since it's way faster. This probably makes more sense to do when we might be ramping up work on the SDK and will want to be able to build quickly and do possible multinode deployments. We can file a separate issue.


charlescrain (approved):
lgtm

Issue #46, opened 2019-08-06
Track `omitempty` issue


safareli:
Opening this to track `omitempty` issue which would result in parse error if tendermint client we have is used against some go implementation which doens't set all props of types generated from proto files.

https://github.com/tendermint/tendermint/issues/3882

martyall:
@charlescrain do the work around with copy 🍝

safareli:
The tendermint PR was merged and changes are in master

IvantheTricourne:
I tested locally with [tendermint v0.32.4](https://github.com/tendermint/tendermint/commit/ff2308b5f4d13e9f88bfc6a82c4f597c8c73c85d) using the e2e test suite in the `hs-abci-example` app. This seems to alleviate the RPC parse exceptions I was getting earlier:

```
uncaught exception: JsonRpcException        

  ParsingException "key \"code\" not present"
```

The responses with the updated `tendermint` look like this:

```json
{
    "result": {
        "response": {
            "log": "",
            "proof": null,
            "height": "0",
            "value": "AAAAAA==",
            "codespace": "",
            "key": "bDVJOiuTeCnJgVw54jr5ZLyE5UMKfcEExwC7wN4rWeM=",
            "code": 0,
            "index": "0",
            "info": ""
        }
    },
    "jsonrpc": "2.0",
    "id": 1064275510610894592
}
```

PR #61, opened 2019-08-09
Example app tests

martyall:
Set up a basic test suite for the example application. finishes #22 

e2e test suite: **blocked**
1. Check the node is alive via `/health`
2. Check that the count was initialized to `0`
3. Set the count to something
4. Query the new count and make sure it got set

application test suite
replicate steps 2 - 4 from above but without using the rpc client. Just plug into the handlers directly. **NOTE** there is no reason to start the node in this case

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,4 +26,4 @@ script:
 # When branch is `master` we run `haskell-stylish` and fail if git working directory becomes dirty
 - if [ "$TRAVIS_BRANCH" == "master" ]; then make stylish && git diff-index --quiet HEAD; fi
 
-- make test
+- make test-libraries
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,3 +27,4 @@ script:
 - if [ "$TRAVIS_BRANCH" == "master" ]; then make stylish && git diff-index --quiet HEAD; fi
 
 - make test-libraries
+- make test-simple-storage-handlers
--- a/hs-abci-example/README.md
+++ b/hs-abci-example/README.md
@@ -8,12 +8,26 @@ allows users to update and query the count.
 There is a `docker-compose.yaml` file in this directory. If you use the `make` command from the project root
 
 ```bash
-> make deploy-simple-storage
+> make deploy-simple-storage-docker
 ```
 
 it will build an image for simple-storage and launch it in a docker network
 with a tendermint-core node. The port for simple-storage is not exposed outside of the docker network --
 of you would like to submit transactions or query state you must do it using the tendermint RPC.
+
+## Running Locally
+Assuming you have a [Tendermint v0.32.2 binary](https://github.com/tendermint/tendermint/releases/tag/v0.32.2) in your path, you can start a tendermint core node with
+
+```bash
+> tendermint init
+> tendermint node --consensus.create_empty_blocks=false
+```
+
+The `--consensus.create_empty_blocks=false` flag is helpful for keeping the logs from being polluted with empty blocks. You can then then start the example application using
+
+```bash
+> make deploy-simple-storage-local
+```
 
 ## Application Messages
 The application uses a protobuf file to define its [transaction messages](https://github.com/f-o-a-m/hs-abci/blob/master/hs-abci-example/protos/simple-storage/messages.proto). Thus if you would like to post transactions to this application via RPC, you will need to first consume
--- a/hs-abci-example/test/SimpleStorage/Test/E2ESpec.hs
+++ b/hs-abci-example/test/SimpleStorage/Test/E2ESpec.hs
@@ -14,16 +14,16 @@ import qualified Network.Tendermint.Client            as RPC
 import           Test.Hspec
 
 
-
 spec :: Spec
 spec = do
-  describe "SimpleStorage E2E" $ do
+  describe "SimpleStorage E2E - via hs-tendermint-client" $ do
 
---    it "Can query /health to make sure the node is alive" $ do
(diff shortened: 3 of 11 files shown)


hs-abci-example/README.md
```diff
@@ -8,13 +8,27 @@ allows users to update and query the count.
 There is a `docker-compose.yaml` file in this directory. If you use the `make` command from the project root
 
 ```bash
-> make deploy-simple-storage
+> make deploy-simple-storage-docker
 ```
 
 it will build an image for simple-storage and launch it in a docker network
 with a tendermint-core node. The port for simple-storage is not exposed outside of the docker network --
 of you would like to submit transactions or query state you must do it using the tendermint RPC.
```
charlescrain: Little typo here, `If you would like to submit trans...`

charlescrain (dismissed):
Looks great. Docs looks good, the single typo isn't enough for me to request changes as it's so minor.

hs-abci-example/test/SimpleStorage/Test/E2ESpec.hs
```diff
@@ -0,0 +1,49 @@
...
+import qualified Network.Tendermint.Client            as RPC
+import           Test.Hspec
+
+
+
+spec :: Spec
+spec = do
+  describe "SimpleStorage E2E" $ do
+
+--    it "Can query /health to make sure the node is alive" $ do
```
charlescrain: It's not clear to me why this is commented out as it looks complete.

charlescrain:
Pretty sure we're being partially blocked by exactly what @safareli pointed out in https://github.com/f-o-a-m/hs-abci/issues/46. It's this and the fact we used the generic FromJSON parser.

Because the protobuf generated types ignore empty values by assigning default values, the types of the missing fields don't need to `Maybe`s. 

(See line 7890 in `hs-abci-types/.stack-work/dist/x86_64-osx/Cabal-2.4.0.1/build/autogen/Proto/Types.hs`).

Essentially a mismatch between the naive generic JSON parsing we use and the more involved parsing that the generated protobuf types use. 

I'm pretty sure the `Data.ProtoLens.Message` class could be used to create a general To/FromJSON instances that comply with the protobufs.

charlescrain:
Haskell needs something like this https://developers.google.com/protocol-buffers/docs/reference/csharp/class/google/protobuf/json-parser

This guy made an encoding library for the proto-lens for ToJSON. Maybe we could look into this for writing the parse?
https://github.com/piesync/proto-lens-json/blob/master/src/Data/ProtoLens/JSON.hs

safareli:
one option is to fork generic{To,From}JSON and add `omitDefaultFields` (similar to `omitNothingFields`) to `Options`. It will omit fields when encoding if they are set to default value and when parsing use values from `def` (Default) if field is ommited from json.

charlescrain:
Most recent commit should complete this. Once reviewed it's should be good to go.

hs-abci-example/test/SimpleStorage/Test/HandlersSpec.hs
```diff
@@ -0,0 +1,62 @@
...
+        .~ encodedUpdateTx
+        )
+      (deliverResp ^. Resp._deliverTxCode) `shouldBe` 0
+      (ResponseQuery queryResp) <- handleQuery
+        (RequestQuery $ defMessage ^. _Unwrapped' & Req._queryPath .~ "count")
+      let foundCount = queryResp ^. Resp._queryValue . to decodeCount
+      foundCount `shouldBe` genCount
+
+encodeCount :: Int32 -> Hex.HexString
+encodeCount 0 = "0x"
```
martyall, line 57: is there a reason we need these special cases?

.travis.yml
```diff
@@ -26,4 +26,4 @@ script:
 # When branch is `master` we run `haskell-stylish` and fail if git working directory becomes dirty
 - if [ "$TRAVIS_BRANCH" == "master" ]; then make stylish && git diff-index --quiet HEAD; fi
 
-- make test
+- make test-libraries
```
martyall, line 29: Can we put your handler test suite in here using some kind of custom stack test command?

martyall (commented):
I think that this looks good, just a few small things. I Also think we can go ahead and merge this with the partial  e2e tests.

hs-abci-example/test/SimpleStorage/Test/HandlersSpec.hs
charlescrain: Yeah the way `HexString` encodes 0 is `0x`

.travis.yml
charlescrain: Like a specific`make test-sample-app-handler`?

martyall: yeah, since we would want to run this particular one in the travis jobs

hs-abci-example/test/SimpleStorage/Test/HandlersSpec.hs
charlescrain: Also to further explain, I was getting parser errors 
``` decodeCount "0x" 
*** Exception: Data.Binary.Get.runGet at position 0: not enough bytes
CallStack (from HasCallStack):
  error, called at libraries/binary/src/Data/Binary/Get.hs:351:5 in binary-0.8.6.0:Data.Binary.Get
```
if I didn't handle the case in `decodeCount`. I added the specific case to `encodeCount` to keep it symmetric.

martyall:
looks good to me! since I opened a PR I can't "approve", gonna merge with admin privs

PR #70, opened 2019-09-19
ABCI Request and Response types with ToJSON and FromJSON instances compliant with Tendermint JSON RPC

charlescrain:


charlescrain:
FYI the brittany formatter changed some of the files, LMK if its not a style you like. 
https://github.com/f-o-a-m/hs-abci/pull/70/files#diff-645530c37356187833b470c48cde8e54R107

hs-abci-types/src/Network/ABCI/Types/Messages/FieldTypes.hs
```diff
@@ -1,3 +1,4 @@
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
```
martyall: we should put this in the `package.yaml` file

martyall (changes requested):

martyall:
How are you invoking this formatter? I don't see any kind of config file or `make` command.

charlescrain:
> How are you invoking this formatter? I don't see any kind of config file or `make` command.

It's the default formatter in hie-vscode

charlescrain:
I went through and fixed the formatting. I also wrote custom parsers where necessary. The parser for HexString now handles, null and parses it as an empty HexString (i.e. `0x`). Since the HexString packaged is defined for use within this package it seemed like a good idea as it saves writing a bunch of custom parsers. Let me know if this doesn't work and I'll dive in and do the custom parsers.

Also, this is not complete, still getting a parser error. when calling tendermint.

charlescrain:
Found another problem with our encoding, we need to use `Base64` strings instead.

martyall (approved):

PR #71, opened 2019-10-25
Polysemy

martyall:
use polysemy for basic module constructions


martyall:
closing in favor of #72

PR #104, opened 2019-11-14
Integrate proto3-suite

### LINKED ISSUE #95: Switch protobuf library to awakesecurity/proto3-suite
https://github.com/awakesecurity/proto3-suite

taken from rocketchat:

(1) we still use protofiles for the Network.ABCI.Types.Messages, and this library can still process those
so there's no difference there
(2) You can still do that for every other type that you want, i.e. ones that maybe you didn't define in your app
(3) Currently you need to write protobuf files by hand in order to make tx message types. This requires you to know protobuf well enough to model all your stuff consistently in proto that you eventually want to represent in haskell

IvantheTricourne:
Closes #95.


hs-abci-examples/nameservice/src/Nameservice/Protogen.hs
```diff
@@ -0,0 +1,73 @@
+{-# LANGUAGE MagicHash #-}
+
+-- @NOTE: ^ possibly the only language extension needed from tutorial
+-- Seems to be the only requirement for generating the .proto string
+
+module Nameservice.Protogen where
```
martyall: This should be moved to a separate directory and labeled as an `executable` in package.yaml

hs-abci-examples/nameservice/src/Nameservice/Modules/Token.hs
```diff
@@ -62,6 +70,16 @@ instance HasCodec Amount where
     encode (Amount b) = cs $ Binary.encode b
     decode = Right . Amount . Binary.decode . cs
 
+-- Address orphans
+instance Primitive Address where
+  encodePrimitive n (Address hx) = Encode.byteString n (Hex.toBytes hx)
+  decodePrimitive = Address . Hex.fromBytes <$> Decode.byteString
+  primType _ = DotProto.Bytes
+instance HasDefault Address where -- @NOTE: deriving generic gets rid of this; see below
```
martyall: is this required? I don't see anything below

hs-abci-examples/nameservice/src/Nameservice/Modules/Nameservice/Types.hs
```diff
@@ -1,42 +1,52 @@
...
   , whoisPrice :: Amount
   } deriving (Eq, Show, Generic)
 
+instance Message Whois
+
+instance Named Whois
+
 instance Binary.Binary Whois
 
 instance HasCodec Whois where
```
martyall, line 49: You should use the binary codecs from protobuf here and remove `Binary` definitions

hs-abci-examples/nameservice/src/Nameservice/Modules/Token.hs
```diff
@@ -62,6 +70,16 @@ instance HasCodec Amount where
     encode (Amount b) = cs $ Binary.encode b
     decode = Right . Amount . Binary.decode . cs
 
+-- Address orphans
```
martyall: if you add orphan instances you should use `{-# OPTIONS_GHC -fno-warn-orphans #-} pragam at the top

martyall (changes requested):

martyall (approved):

PR #106, opened 2019-11-19
Martin/query client

martyall:
implement client interpretation for storage querying to make testing slightly easier.


hs-abci-examples/simple-storage/test/SimpleStorage/Test/E2ESpec.hs
```diff
@@ -30,14 +38,16 @@ spec = do
...
-      let foundCount = queryResp ^. Resp._queryValue . to decodeCount
+      let queryReq = QueryArgs
+            { queryArgsData = SS.CountKey
+            , queryArgsHeight = 0
+            , queryArgsProve = False
+            }
+      SS.Count foundCount <- do
+        eResp <- runQueryRunner $ getCount queryReq
+        case eResp of
+          Left e                   -> error $ "Parse Error: " <> show e
```
martyall: I don't think this can actually happen

PR #114, opened 2019-11-20
Nameservice tests

### LINKED ISSUE #107: Nameservice tests
we need to write e2e tests for the nameservice app that test basic getting, setting, and buying of names. 

Since we're still waiting on a tendermint 32.4 docker image, these should not run on CI yet. They should pass locally

IvantheTricourne:
Closes #107


hs-abci-examples/nameservice/test/Nameservice/Test/E2ESpec.hs
```diff
@@ -0,0 +1,244 @@
...
+        let nope = Name "nope"
+            queryReq = defaultReqWithData nope
+        ClientResponse{ clientResponseData = emptyWhois
+                      , clientResponseRaw
+                      } <- runRPC $ getWhois queryReq
+        let queryRespCode = clientResponseRaw ^. Response._queryCode
+        -- storage failure
+        queryRespCode `shouldBe` 1
+        -- empty whois (defaults)
+        whoisPrice emptyWhois `shouldBe` 0
```
martyall, line 99: Am I understanding that the protobuf messages can be decoded from an empty bytestring, similar to how the EVM returns empty structs even if they aren't found in a mapping?

hs-abci-examples/nameservice/test/Nameservice/Test/E2ESpec.hs
```diff
@@ -0,0 +1,244 @@
...
+
+spec :: Spec
+spec = do
+  let satoshi = Name "satoshi"
+      addr1 = userAddress user1
+      privateKey1 = userPrivKey user1
+      addr2 = userAddress user2
+      privateKey2 = userPrivKey user2
+
+  beforeAll (do faucetAccount user1; faucetAccount user2) $
```
martyall, line 60: nice before step

hs-abci-sdk/src/Tendermint/SDK/Types/Transaction.hs
```diff
@@ -38,12 +39,14 @@ data RawTransaction = RawTransaction
   , rawTransactionSignature :: ByteString
   } deriving Generic
 
+-- basically also a message type
 instance Serialize.Serialize RawTransaction
 
 parseRawTransaction
```
martyall, line 45: This use of Data.Serialize is a fine stopgap, but go ahead and make an issue to write a protobuf file for this in the sdk itself so we can get away from it.

martyall (changes requested):
This looks great, just a few remarks. But we should also go ahead and test event emissions here. It's not necessary all the time, let's pick one place. Note that you can use https://github.com/f-o-a-m/hs-abci/blob/b28f051ee163053834b6a1b1491e4136f18ddbc0/hs-abci-sdk/src/Tendermint/SDK/Events.hs#L52-L68 . It has a unit test here https://github.com/f-o-a-m/hs-abci/blob/b28f051ee163053834b6a1b1491e4136f18ddbc0/hs-abci-sdk/test/Tendermint/SDK/Test/EventSpec.hs that proves that it should work, but let's hope so.

hs-abci-examples/nameservice/test/Nameservice/Test/E2ESpec.hs
IvantheTricourne: Yeah, I think the proto3-suite decoding library is meant to behave that way. Each field of a decoded message is assumed to have a `HasDefault` instance.

hs-abci-sdk/src/Tendermint/SDK/Types/Transaction.hs
IvantheTricourne: See #117

martyall (approved):

PR #116, opened 2019-11-20
A more thorough CI pipeline

### LINKED ISSUE #111: Make a real CI pipeline for this project
I don't care if it's travis or jenkins. This real CI pipeline should be able to 

1. create branch and merge jobs for each PR
2. build the projects
3. Test all haskell unit tests in all stack projects
4. test the rpc client tests against kvstore app baked into tendermint, see the [README](https://github.com/f-o-a-m/hs-abci/blob/master/hs-tendermint-client/README.md).
5. Run e2e test suite for simple storage against the docker-compose network
6. Run the e2e test suite for nameservice against the docker-compose network

it would be nice if the tests run in parallel.

By docker-compose network, I mean the one in the docker compose file. For example, here is the one for simple storage [docker-compose file](https://github.com/f-o-a-m/hs-abci/blob/master/hs-abci-examples/simple-storage/docker-compose.yaml)

There is a non-functioning dockerfile for each example app. It is an outstanding issue to fix it. I don't care if those e2e tests run the example app in docker, or if they run the app locally against the docker network that it depends on.

NOTE: The tests requiring a tendermint core docker image must wait until they push a 0.32.4 tag to dockerhub, or we must make this image ourselves and host it from our docker repo.

We can continue to run a merge travis job for hlint and stylish when the target branch is master.

Eventually there will probably be a nix component to this, but only if one of the future consultants can do most of the heavy lifting there.

iostat:
Fixes #111 (mostly)

1. Done by Travis automagically
2. Prerequisite for running the tests
3. Done via `make test-libraries`
4. Created a `docker-compose.yaml` similar to the one used for E2E, which builds an `abci-cli` container, then runs `make test-kv-store`
5. Created a Dockerfile that builds the entire the `hs-abci` suite in Docker, which E2E test `docker-compose.yaml`s can fall back to if necessary. The is also a `Dockerfile.prebake`, meant to be used in conjunction with the `docker-test-prebake` make target. This target builds all binaries locally (allowing us to take advantage of Travis caching `.stack`) before copying them into a container image (specified in `Dockerfile.prebake`) which is behaviorally identical to the one in the `Dockerfile` (`FROM haskell:8`, `protoc`+`libsecp256k1`, and `hs-abci` binaries in `/usr/local/bin`). Because Travis is running Linux, the binaries its stack compiles would be compatible with the Docker containers we're making and we can just copy them wholesale instead of building inside Docker which takes upwards of an hour. Note that this optimization won't work if you're not on a Linux machine. To tie all this together, the E2E tests' `docker-compose.yaml` specify an image of `hs-abci:test` which the `docker-test-prebake` make target builds. If that image is unavailable for whatever reason (i.e., one didn't run `make docker-test-prebake` beforehand), the `build` section inside `docker-compose.yaml` tells `docker-compose` to fall back to building the image using the "slow", full `Dockerfile`.
6. nameserver tests didn't exist yet, but to create them -- assuming they're structured similarly to the `simple-storage` tests -- is to copypasta `docker-compose.yaml` from `simple-storage`, change the `entrypoint` line to use `nameserver` instead of `simple-storage`, and copypasta the job in `travis.yml` replacing all instances of `simple-storage` with `nameserver` (and changing the `name:` accordingly).
7. (styish/hlint) I moved the `if TRAVIS_BRANCH == master` line from the original script into its own sub-job which only runs if the branch is master. Not sure if you wanted me to run `hlint` as well as it wasn't there in the first place.

Additionally, you'll find that if you hit the `Details` link in the CI reports at the bottom of this PR, you'll see there are four test jobs in the `Branch` build, whereas if you look at the details for the `Pull Request` build, you'll see the additional job for `stylish` (as this is going to be merged into master)


iostat:
The current failures we're seeing here appear to be an issue with the tests themselves -- perhaps a consequence of using Tendermint `v0.32.8`?

```
RPC Request
{
    "jsonrpc": "2.0",
    "params": {
        "height": "0",
        "path": "/simple_storage/count/count",
        "data": "6c35493a2b937829c9815c39e23af964bc84e5430a7dc104c700bbc0de2b59e3",
        "prove": false
    },
    "method": "abci_query",
    "id": 143685140804222960
}
RPC Response
{
    "result": {
        "response": {
            "log": "Path Not Found",
            "proof": null,
            "height": "0",
            "value": null,
            "codespace": "/simple_storage/count/count",
            "key": null,
            "code": 1,
            "index": "0",
            "info": ""
        }
    },
    "jsonrpc": "2.0",
    "id": 143685140804222960
}
    Can query the count and make sure its initialized to 0 FAILED [1]
Failures:
  src/Tendermint/SDK/Query/Client.hs:59:29: 
  1) SimpleStorage.Test.E2E, SimpleStorage E2E - via hs-tendermint-client, Can query the count and make sure its initialized to 0
       uncaught exception: ErrorCall
       Impossible parse error: too few bytes
       From:	demandInput
       
       
       CallStack (from HasCallStack):
         error, called at src/Tendermint/SDK/Query/Client.hs:59:29 in hs-abci-sdk-0.1.0.0-BzXwLwijEdxJsr1Xk3M7ER:Tendermint.SDK.Query.Client
  To rerun use: --match "/SimpleStorage.Test.E2E/SimpleStorage E2E - via hs-tendermint-client/Can query the count and make sure its initialized to 0/"
Randomized with seed 1567128490
Finished in 0.9371 seconds
5 examples, 1 failure
```

@blinky3713 this is outside my forte

martyall (approved):
🎬

PR #118, opened 2019-11-21
Token modules

### LINKED ISSUE #115: Make module level transaction messages for the "token" module
make `transfer` / `faucet` messages following the pattern here https://github.com/f-o-a-m/hs-abci/blob/module-defs/hs-abci-examples/nameservice/src/Nameservice/Modules/Nameservice/Messages.hs

Make router for those messages similar to Nameservice

Also go ahead and split the Token.hs file into smaller compenets like Nameservice

IvantheTricourne:
Closes #115.


IvantheTricourne:
WIP. Needs test for token routes. Will merge after #114.

martyall (approved):

PR #125, opened 2019-11-22
added gas metering and tests

### LINKED ISSUE #121: Implement Gas Metering
We would like to be able to perform gas metering for certain operations, e.g. get and set in a database, or other custom actions defined in a modules effect systems. This metering should be done on interpretation in the module's eval function

This will likely require adding gas fields to the raw transaciton

martyall:
fixes #121 

I'm reluctant to actually add gas metering to anything at all right now, but this at least gives a sketch on how to do it for basically any action

Even if we were only going to meter database operations, probably the cost of these things is taken from some config that we don't know and don't want to deal with for a while.

Maybe we can just merge this and keep it around for future discussions.

### DIFF
--- a/hs-abci-sdk/package.yaml
+++ b/hs-abci-sdk/package.yaml
@@ -109,6 +109,7 @@ library:
   - Tendermint.SDK.Types.Address
   - Tendermint.SDK.TxRouter
   - Tendermint.SDK.Module
+  - Tendermint.SDK.Gas
 
 tests:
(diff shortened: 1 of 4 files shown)


charlescrain (approved):
LGTM. Just mentioning this for down the line, here is the cosmos default config:
https://github.com/cosmos/cosmos-sdk/blob/master/store/types/gas.go#L153-L163

PR #127, opened 2019-11-25
Don't decode query data into defaults

### LINKED ISSUE #124: make sure that storage lookup misses are reported back as Nothing
It seems like the nameservice tests show that when using proto3-suite if you're trying to decode an empty object it will just return the `Default`. This is basically what the EVM is currently doing and it's extremely annoying, we should make sure that it is returning `Nothing`.

IvantheTricourne:
Closes #124.

### DIFF
--- a/hs-abci-examples/nameservice/test/Nameservice/Test/E2ESpec.hs
+++ b/hs-abci-examples/nameservice/test/Nameservice/Test/E2ESpec.hs
@@ -65,7 +65,7 @@ spec = do
 
       it "Can query account balances" $ do
         let queryReq = defaultReqWithData addr1
-        ClientResponse{clientResponseData = Just foundAmount} <- runRPC $ getBalance queryReq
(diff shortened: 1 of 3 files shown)


hs-abci-examples/nameservice/test/Nameservice/Test/E2ESpec.hs
```diff
@@ -65,7 +65,7 @@ spec = do
 
       it "Can query account balances" $ do
         let queryReq = defaultReqWithData addr1
-        ClientResponse{clientResponseData = foundAmount} <- runRPC $ getBalance queryReq
+        ClientResponse{clientResponseData = Just foundAmount} <- runRPC $ getBalance queryReq
```
martyall: You should be checking that the response code is `0` before asserting `Just` here. Otherwise you will get an uninformative runtime error message. Similar witht he other tests.

martyall (changes requested):
check the `0` response code

martyall (approved):

PR #135, opened 2019-12-02
Implement default codec for messages

### LINKED ISSUE #134: implement default HasCodec instance for proto3-suite messages
Any type that has a `Proto3.Suite.Message` instance should have a default `HasCodec` instance (i.e., using the codecs from `Proto3.Suite`).

This alleviates the need to repeatedly do [this](https://github.com/f-o-a-m/hs-abci/blob/master/hs-abci-examples/nameservice/src/Nameservice/Modules/Nameservice/Messages.hs#L40-L42)

IvantheTricourne:
Closes #134


martyall:
in the end I don't think we're going to do this

PR #99, opened 2019-11-13
intial metrics setup

charlescrain:

PR #144, opened 2019-12-10
impl metrics effect

### LINKED ISSUE #79: investigate adding metrics middleware with datadog backend
- [x] make a middleware for server metrics, just measuring response times and counters for messages.

- [ ] investigate making log scribes with ES and Kibana

The end result here is that we can run a node for "a long time" (enough time to accrue data), and collect metrics on it's respnonse times and message counts.

Is there a dockerized setup with a dashboard for viewing metrics? Alternatively there's the "official" dashboard tied to a datadog account. Either way we want to view them.

### LINKED ISSUE #138: Sketch Metrics Effect
We should add a `Metrics` effect to the SDK. This effect should be able to update counts (number of times a message type appears) and the amount of time an action takes. Additionally, we would want to be able to log metrics (similar to console logging):

```haskell
data Metrics m a where
  -- | Increments the count of a specific message
  IncCount :: CountName -> Metrics m ()
  -- | Times an action
  WithTimer :: m a -> Metrics m (a, NominalDiffTime)
```

UPDATE: metrics won't have logging. Interpreter will be responsible for taking summands to a [prometheus](https://github.com/bitnomial/prometheus). Other summands should be added to support this interface (i.e., `getCount`, `observeHistogram`, `getHistogram`).

IvantheTricourne:
Closes #99.
Closes #79.
Closes #138.


IvantheTricourne:
TODO: Not sure how to get populate the metrics server

hs-abci-extra/src/Network/ABCI/Server/Middleware/MetricsLogger.hs
```diff
@@ -0,0 +1,145 @@
...
+  scribe <- liftIO $ mkDatadogScribe datadogScribeSettings (permitItem InfoS) V0
+  le <- liftIO (registerScribe "datadog" scribe defaultScribeSettings
+        =<< initLogEnv "ABCI" "production")
+  let ns = "Server"
+  mvarReqC <- liftIO $ newMVar Map.empty
+  pure $ mkMetricsLogger mvarReqC le ns
+
+-- mkMetricsLogDatadog :: (MonadIO m) => m (Middleware m)
+-- mkMetricsLogDatadog = do
+--   let apiKey = APIKey undefined
```
IvantheTricourne: Unsure how to address this. Maybe an env var?

hs-abci-examples/nameservice/src/Nameservice/Server.hs
```diff
@@ -6,29 +6,36 @@ import           Nameservice.Application                       (AppConfig (..),
...
+      nat = runCoreEffs baseAppContext
       application = createIOApp nat $ makeApp handlersContext
   serveApp =<< hookInMiddleware application
   where
     mkMiddleware :: IO (Middleware IO)
     mkMiddleware = do
       reqLogger <- ReqLogger.mkLogStdoutDev
       resLogger <- ResLogger.mkLogStdoutDev
+      -- metLogger <- MetLogger.mkMetricsLogDatadogLocal 10516 -- local port for log collection (TDP)
+      --metLogger <- MetLogger.mkMetricsLogDatadog -- direct datadog
```
IvantheTricourne: Perhaps a local scribe is fine?

hs-abci-sdk/src/Tendermint/SDK/BaseApp/Metrics.hs
```diff
@@ -0,0 +1,36 @@
...
+instance IsString HistogramName where
+  fromString s = HistogramName (fromString s) mempty defaultBuckets
+    where defaultBuckets = [0.0001, 0.001, 0.01, 0.1, 0.25, 0.5, 0.75, 1, 10, 100]
+
+-- @NOTE: might need a NewCount/NewHistogram summand to clean up some usage quirks
+data Metrics m a where
+  -- | Increments the count of a specific message
+  IncCount :: CountName -> Metrics m ()
+  -- | Update a histogram
+  ObserveHistogram :: HistogramName -> Double -> Metrics m ()
```
martyall: What is the target use case for this? Usually this would make sense to support a timer, but we already have a command for timer that doesn't need to use this to be interpreted

hs-abci-sdk/src/Tendermint/SDK/BaseApp/Metrics/Prometheus.hs
```diff
@@ -0,0 +1,187 @@
...
+            liftIO $ Histogram.observe val hist
+            pure histMap
+      pureT ()
+
+    WithTimer histName action -> do
+      start <- liftIO $ getCurrentTime
+      a <- runT action
+      end <- liftIO $ getCurrentTime
+      let time = fromRational . (* 1000.0) . toRational $ (end `diffUTCTime` start)
+      evalMetrics state (observeHistogram histName time)
```
martyall: To me this is further evidence that observeHistogram should just be a function used in the interpreter written in terms of `IO`, or something else outside the `Metrics` DSL

hs-abci-examples/nameservice/src/Nameservice/Modules/Nameservice/Router.hs
```diff
@@ -20,6 +23,18 @@ router
 router (RoutedTx Tx{txMsg}) =
   let Msg{msgData} = txMsg
   in case msgData of
-       NSetName msg    -> setName msg
-       NBuyName msg    -> buyName msg
-       NDeleteName msg -> deleteName msg
+       NSetName msg    -> do
+         incCount "count_set"
+         observeHistogram "histogram_set" 0.0
```
martyall: what are these uses of `observeHistogram` attempting to record?

hs-abci-examples/nameservice/src/Nameservice/Server.hs
martyall: Wouldn't you rather use environment variables so we could publish to the same datadog account that the metrics dsl is using?

hs-abci-extra/src/Network/ABCI/Server/Middleware/MetricsLogger.hs
martyall: why not take it as an arg and use an environment variable whenever it gets used in application code to create this arg

hs-abci-sdk/src/Tendermint/SDK/BaseApp/Metrics/Prometheus.hs
```diff
@@ -0,0 +1,187 @@
...
+  }
+
+instance IsString MetricIdentifier where
+  fromString s = MetricIdentifier (fromString s) mempty mempty
+
+--------------------------------------------------------------------------------
+-- setup @TODO: add error handling
+--------------------------------------------------------------------------------
+
+getEnvVarBoolWithDefault :: MonadIO m => String -> Bool -> m Bool
```
martyall: This shouldn't really be here. You should only use environment variables in application code.  for example this

```
  shouldStart <- getEnvVarBoolWithDefault "STATS_ENABLED" True
  if not shouldStart
    then return Nothing
    else do
      bindPort <- readEnvVarWithDefault "STATS_PORT" 9200

```

means that this function should take a `Maybe Port` and figure out what to do based on `Nothing | Just Port` and then shove the `Maybe Port` as part of the metrics config. Then in application code you can use an EnvVar to discover the value.

martyall (changes requested):

hs-abci-sdk/src/Tendermint/SDK/BaseApp/Metrics.hs
IvantheTricourne: Valid point.

I don't remember if i've noted it somewhere, but the usage of these summands are also to init a histogram, since the act of "adding a hist" to a prometheus registry doesn't allow to set the starting value. You can see this in the tests where `observeHistogram` is passed `0.0`.

hs-abci-sdk/src/Tendermint/SDK/BaseApp/Metrics/Prometheus.hs
IvantheTricourne: Yeah that makes sense.

hs-abci-sdk/src/Tendermint/SDK/BaseApp/Metrics.hs
IvantheTricourne: An alternative to this is changing `observeHistogram` to an init hist function and using `observeHistogram` as just an `IO` operation.

hs-abci-examples/nameservice/src/Nameservice/Modules/Nameservice/Router.hs
IvantheTricourne: This is how you init a histogram. The value here (i.e., `0.0`) is just a dummy value.

hs-abci-sdk/src/Tendermint/SDK/BaseApp/Metrics/Prometheus.hs
IvantheTricourne: Does that have to be part of the BaseApp context similar to logging? Or can it just live inside of the AppConfig?

hs-abci-examples/nameservice/app/Main.hs
```diff
@@ -1,19 +1,30 @@
...
+import           Tendermint.SDK.BaseApp.Logger.Katip       (LogConfig (..),
+                                                            mkLogConfig)
+import           Tendermint.SDK.BaseApp.Metrics.Prometheus (MetricsConfig (..),
+                                                            mkMetricsConfig)
+import qualified Text.Read                                 as T
 
 main :: IO ()
 main = do
+  metCfg@MetricsConfig{..} <- mkMetricsConfig
+  mApiKey <- lookupEnv "DD_API_KEY"
```
martyall, line 19: we should put defaults in the Makefile when possible, and otherwise do something like this https://github.com/f-o-a-m/foam.tcr/blob/master/Makefile#L291 so we get an explicit list of all required vars and can see the bindings at startup

hs-abci-examples/nameservice/src/Nameservice/Modules/Nameservice/Router.hs
```diff
@@ -20,6 +21,15 @@ router
 router (RoutedTx Tx{txMsg}) =
   let Msg{msgData} = txMsg
   in case msgData of
-       NSetName msg    -> setName msg
-       NBuyName msg    -> buyName msg
-       NDeleteName msg -> deleteName msg
+       NSetName msg    -> do
```
martyall, line 24: these functions already return `()`, so there's no need for the wildcard binding and `return ()`

hs-abci-extra/src/Network/ABCI/Server/Middleware/MetricsLogger.hs
```diff
@@ -0,0 +1,152 @@
...
+
+---------------------------------------------------------------------------
+-- mkMetricsLogStdout
+---------------------------------------------------------------------------
+-- | Creates a production Metrics logger as middleware for ABCI Server.
+-- Uses lowest possible verbosity, however, with the current minimal metrics
+-- verbosity has no effect on which metrics logged.
+mkMetricsLogStdout :: (MonadIO m) => m (Middleware m)
+mkMetricsLogStdout = do
+  handleScribe <- liftIO $ mkHandleScribe ColorIfTerminal stdout (permitItem InfoS) V0
```
martyall: we can factor out the `liftIO` to before the `do` statement 

liftIO (m >>= f) = liftIO m >>= (liftIO . f)


https://hackage.haskell.org/package/transformers-0.4.2.0/docs/Control-Monad-IO-Class.html

hs-abci-extra/src/Network/ABCI/Server/Middleware/MetricsLogger.hs
```diff
@@ -0,0 +1,152 @@
...
+        =<< initLogEnv "ABCI" "production")
+  let ns = "Server"
+  mvarReqC <- liftIO $ newMVar Map.empty
+  pure $ mkMetricsLogger mvarReqC le ns
+
+---------------------------------------------------------------------------
+-- mkMetricsLogDatadog
+---------------------------------------------------------------------------
+
+mkMetricsLogDatadogLocal :: (MonadIO m) => Integer -> m (Middleware m)
```
martyall, line 46: same here with the `liftIO`

hs-abci-extra/src/Network/ABCI/Server/Middleware/MetricsLogger.hs
```diff
@@ -0,0 +1,152 @@
...
+  datadogScribeSettings <- liftIO $ mkDatadogScribeSettings (localAgentConnectionParams (fromInteger port)) NoAuthLocal
+  scribe <- liftIO $ mkDatadogScribe datadogScribeSettings (permitItem InfoS) V0
+  le <- liftIO (registerScribe "datadog" scribe defaultScribeSettings
+        =<< initLogEnv "ABCI" "production")
+  let ns = "Server"
+  mvarReqC <- liftIO $ newMVar Map.empty
+  pure $ mkMetricsLogger mvarReqC le ns
+
+newtype ApiKey = ApiKey Text
+mkMetricsLogDatadog :: (MonadIO m) => Maybe ApiKey -> m (Middleware m)
```
martyall, line 57: same here with `liftIO`

hs-abci-examples/nameservice/src/Nameservice/Modules/Nameservice/Router.hs
```diff
@@ -20,6 +21,15 @@ router
...
+         incCount "count_set"
+         _ <- withTimer "histogram_set" $ setName msg
```
martyall: is there a reason to prefix the histogram with "histogram" or the counter with "count"?

hs-abci-sdk/src/Tendermint/SDK/BaseApp/Metrics/Prometheus.hs
```diff
@@ -0,0 +1,205 @@
...
+          Nothing -> do
+            newCtr <- liftIO $
+              Registry.registerCounter cMetricIdName metricIdLabels metricsRegistry
+            let newCounterMap = insert cid newCtr counterMap
+            liftIO $ Counter.inc newCtr
+            pure newCounterMap
+          Just ctr -> do
+            liftIO $ Counter.inc ctr
+            pure counterMap
+      pureT ()
```
martyall, line 79: is this actually needed given that `modifyMVar_` returns `()`?

hs-abci-sdk/src/Tendermint/SDK/BaseApp/Metrics/Prometheus.hs
```diff
@@ -0,0 +1,205 @@
...
+
+    -- | Updates a histogram with the time it takes to do an action
+    -- | If histogram doesn't exist, creates a new one and observes it.
+    WithTimer histName action -> do
+      start <- liftIO $ getCurrentTime
+      a <- runT action
+      end <- liftIO $ getCurrentTime
+      let time = fromRational . (* 1000.0) . toRational $ (end `diffUTCTime` start)
+      observeHistogram state histName time
+      _ <- raise $ evalMetrics state a
```
martyall, line 89: why are we discarding this value? is it always `()`, if so, then do you actually need the `pureT ()` ?

hs-abci-sdk/src/Tendermint/SDK/BaseApp/Metrics/Prometheus.hs
```diff
@@ -0,0 +1,205 @@
...
+      pureT ()
+    )
+
+-- | Updates a histogram with an observed value
+observeHistogram :: MonadIO m => MetricsState -> HistogramName -> Double -> m ()
```
martyall, line 94: use `liftIO` before the `do` (monad transformer laws)

hs-abci-sdk/src/Tendermint/SDK/BaseApp/Metrics/Prometheus.hs
```diff
@@ -0,0 +1,205 @@
...
+
+runMetricsServer
+  :: MonadIO m
+  => Maybe MetricsConfig
+  -> m ()
+runMetricsServer mMetCfg = liftIO $ do
+  unless (isNothing mMetCfg) $ do
+    let MetricsConfig{..} = fromJust mMetCfg
+        MetricsState{..} = metricsState
+    unless (isNothing metricsPort) $ do
```
martyall: use a `case` statement to avoid `fromJust`

martyall (changes requested):
overall looks good, mostly stylistic things

hs-abci-examples/nameservice/src/Nameservice/Modules/Nameservice/Router.hs
IvantheTricourne: right forgot to change it

hs-abci-examples/nameservice/src/Nameservice/Modules/Nameservice/Router.hs
IvantheTricourne: Nah. It can be named anything. I tried looking for some sort of convention elsewhere, but I ended up just settling for these names.

hs-abci-sdk/src/Tendermint/SDK/BaseApp/Metrics/Prometheus.hs
IvantheTricourne: Yeah that's the only way i ever got it to type check properly.

hs-abci-sdk/src/Tendermint/SDK/BaseApp/Metrics/Prometheus.hs
IvantheTricourne: I never gotten it to typecheck otherwise. The result of `evalMetrics state a` (at least for our cases) is almost always `()`, but I don't think the type checker knows that given the type of the `withTimer` summand is over a general `m a`.

hs-abci-extra/src/Network/ABCI/Server/Middleware/MetricsLogger.hs
IvantheTricourne: seems useful for `hlint`

martyall (approved):

PR #147, opened 2019-12-12
remove cereal, make auth a real module

### LINKED ISSUE #139: Remove dependency on cereal

martyall:
fixes #139

### DIFF
--- a/hs-abci-sdk/package.yaml
+++ b/hs-abci-sdk/package.yaml
@@ -52,8 +52,6 @@ dependencies:
 - avl-auth
 - base >= 4.7 && < 5
 - bytestring
-- cereal
-- cereal-text
 - containers
 - cryptonite
 - data-default-class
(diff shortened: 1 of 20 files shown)


hs-abci-sdk/src/Tendermint/SDK/Crypto.hs
```diff
@@ -79,7 +79,14 @@ instance RecoverableSignatureSchema Secp256k1 where
     recover _ sig dig = Secp256k1.recover sig (msgFromSHA256 dig)
     -- NOTE: I think the use of Data.Serialize is harmless here, because it basically
```
IvantheTricourne: Probably don't need this comment anymore then?

hs-abci-sdk/package.yaml
```diff
@@ -150,12 +150,12 @@ tests:
     - -with-rtsopts=-N
 
     dependencies:
+      - cereal
```
IvantheTricourne, line 153: So cereal is only used for some tests?

IvantheTricourne (commented):
Seems legit. Does simple storage really need cereal still?

martyall:
I think it's fine for tests to use `cereal`. Simple storage does because it's really convenient, and is effectively a test, idk. In order to remove it I need to put it's storage types in a `.proto` file, which in this case was just an `Int32`, it seemed lame.

charlescrain (approved):

PR #148, opened 2019-12-12
Separate tx checker

### LINKED ISSUE #140: Be able to choose to run tx or not in checkTx
RIght now the checkTx message will automatically run the transaction and rollback _if the `validateMessage` function returns `Success`_.

ideally you could decide what the `checkTx` requirements are, maybe it should just run the `validateMessage`.

martyall:
fixes #140

### DIFF
--- a/hs-abci-examples/nameservice/test/Nameservice/Test/E2ESpec.hs
+++ b/hs-abci-examples/nameservice/test/Nameservice/Test/E2ESpec.hs
@@ -164,7 +164,7 @@ spec = do
         -- try to buy at a lower price
         let msg = TypedMessage "BuyName" (encode $ BuyName 100 satoshi "hello (again) world" addr1)
             rawTx = mkSignedRawTransactionWithRoute "nameservice" privateKey1 msg
-        ensureCheckAndDeliverResponseCodes (0,1) rawTx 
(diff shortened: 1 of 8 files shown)


hs-abci-examples/nameservice/test/Nameservice/Test/E2ESpec.hs
```diff
@@ -118,8 +118,7 @@ spec = do
         -- try to set a name without being the owner
         let msg = TypedMessage "SetName" (encode $ SetName satoshi addr2 "goodbye to a world")
             rawTx = mkSignedRawTransactionWithRoute "nameservice" privateKey2 msg
-        checkResp <- getCheckTxResponse rawTx
-        ensureCheckResponseCode checkResp 2
+        ensureCheckAndDeliverResponseCodes (0,2) rawTx
```
IvantheTricourne, line 121: I guess this translates to: _message is well formed but (in this context) not semantically._

martyall: yes, it means that the tx will make it past the mempool when it passes check tx when the tx is run in a block it will produce an error code 2

IvantheTricourne (approved):
LGTM.

PR #151, opened 2019-12-13
Hs abci test utils

### LINKED ISSUE #143: create hs-abci-test-utils package
- [x] factor out all the duplication from nameservice and simple storage utils for managing signing, utils, and expected tx results
- [x] move the client generator from SDK into here, and remove the sdk depenency on the tendermint-client-lib
- [x] move the `FromEvent` class from the `Events` module into here, since again there's no reason the SDK needs to decode events

IvantheTricourne:
Closes #143


hs-abci-test-utils/src/Tendermint/Utils/Events.hs
```diff
@@ -0,0 +1,33 @@
+module Tendermint.Utils.Events where
```
IvantheTricourne, line 1: This module gives off different vibes than the rest in here. This one is actually used in application code (i.e., `nameservice/types`). The rest are only used in tests.

martyall (approved):

PR #162, opened 2019-12-23
Impl Logger AddContext summand

### LINKED ISSUE #161: Add logger context summand
This will allow the logging effect to add contexts (e.g. katip contexts) to logs.

IvantheTricourne:
Closes #161.


hs-abci-extra/src/Network/ABCI/Server/Middleware/ResponseLogger.hs
```diff
@@ -26,33 +22,7 @@ instance ToObject (Loggable (Response (t :: MessageType))) where
...
--- | Creates a request logger as middleware for ABCI requests.
--- Uses highest possible verbosity.
-mkLogStdoutDev :: (MonadIO m) => m (Middleware m)
-mkLogStdoutDev = do
-  handleScribe <- liftIO $ mkHandleScribe ColorIfTerminal stdout (permitItem DebugS) V3
-  le <- liftIO (registerScribe "stdout" handleScribe defaultScribeSettings
-        =<< initLogEnv "ABCI" "development")
-  let ns = "Server"
-  pure $ mkResponseLogger le ns
+  payloadKeys _ _  = SomeKeys ["type", "event_type"]
```
martyall: I don't think `event_type` appears as a key in any place where this is used.

hs-abci-extra/src/Network/ABCI/Server/Middleware/RequestLogger.hs
```diff
@@ -23,7 +23,7 @@ instance ToObject (Loggable (Request (t :: MessageType))) where
 
 instance LogItem (Loggable (Request (t :: MessageType))) where
   payloadKeys V3 _ = AllKeys
-  payloadKeys _ _  = SomeKeys ["type"]
+  payloadKeys _ _  = SomeKeys ["type", "event_type"]
```
martyall: I don't think `event_type` appears as a key in any place where this is used.

hs-abci-sdk/src/Tendermint/SDK/BaseApp/Logger.hs
```diff
@@ -3,17 +3,43 @@
...
+data Verbosity = V0 | V1 | V2 | V3
+
+-- | Class for selecting object keys for contextual logging
+class Select a where
+  select :: Verbosity -> a -> LogSelect
+  default select :: Verbosity -> a -> LogSelect
+  select _ _ = All
+
+-- | Special event wrapper to add contextual event_type info
+newtype ContextEvent t = ContextEvent t
```
martyall: This maybe should live in the `BaseApp.Events` module

hs-abci-examples/nameservice/src/Nameservice/Modules/Nameservice/Keeper.hs
```diff
@@ -63,14 +63,17 @@ setName SetName{..} = do
...
-             , nameRemappedNewValue = setNameValue
-             , nameRemappedOldValue = whoisValue
-             }
+          let event = NameRemapped
+                { nameRemappedName = setNameName
+                , nameRemappedNewValue = setNameValue
+                , nameRemappedOldValue = whoisValue
+                }
+          BaseApp.emit event
+          BaseApp.addContext (BaseApp.ContextEvent event) $
```
martyall: this same expression appears multiple times, it's better to write some function like 

```haskell
logEvent
  :: Select e
  => Member Logging r
  => e
  -> Sem r ()
logEvent e =  BaseApp.addContext (BaseApp.ContextEvent e) $
  BaseApp.log BaseApp.Debug (cs $ BaseApp.makeEventType (Proxy :: Proxy e))
```

This probably lives in BaseApp.Events

martyall: It's also probably an `Info` level log

hs-abci-examples/nameservice/README.md
```diff
@@ -1,23 +1,147 @@
 # nameservice
 
-## Metrics
+## Metrics via Prometheus
```
martyall, line 3: Before commenting on any of this, is there any way to have kibana pull metrics from prometheus in the same way datadog does. This would be good to fix some confusing asymmetry between server metrics and application metrics, and would also be easier to explain that you can now use either Kibana or datadog to view anything you want, or both.

martyall (approved):

hs-abci-examples/nameservice/README.md
IvantheTricourne: Good question. I’ll investigate.

IvantheTricourne: @blinky3713 

It seems [MetricBeat](https://www.elastic.co/products/beats/metricbeat) is an option from the ELK folks that can [scrape prometheus exporters and servers](https://www.elastic.co/guide/en/beats/metricbeat/current/metricbeat-metricset-prometheus-collector.html). Given we're already using ELK stack, this seems to be a pretty good option.

Similar to the datadog agent, it also has a [docker-compose](https://github.com/shazChaudhry/docker-elastic/blob/master/metricbeat-docker-compose.yml).

hs-abci-extra/src/Network/ABCI/Server/Middleware/ResponseLogger.hs
IvantheTricourne: u rite. I just thought this was needed.

hs-abci-sdk/src/Tendermint/SDK/BaseApp/Events.hs
```diff
@@ -121,6 +125,35 @@ emit
...
+logEvent
+  :: forall e r.
+     (A.ToJSON e, ToEvent e, Log.Select e)
+  => Member Log.Logger r
+  => e
+  -> Sem r ()
+logEvent event = Log.addContext (ContextEvent event) $
+  Log.log Log.Info (cs $ makeEventType (Proxy :: Proxy e))
+
+emitAndLogEvent
```
martyall: I don't think we need this in the library, it's simple enough to use both functions when both are desired

martyall: I don't think we should put this one in the library, it's simple enough to call both functions

martyall (changes requested):
There is a problem with `checkTx` / `deliverTx` with how this is now. Namely, if `checkTx` is actually running transactions without committing state, then it's going to log the same event to ES twice. We need to set a context for `check` versus `deliver`, and probably need to set context for the transaction hash as well so things are easily searchable.

 I think it may make more sense to leave this unsolved here and instead fix it in #167 since transactions are further isolated in that PR

martyall (approved):

PR #168, opened 2020-01-02
Nameservice interact

IvantheTricourne:
Closes #157

PR #179, opened 2020-01-06
Async nameservice interact

### LINKED ISSUE #157: make script for interacting with the nameservice app
this script should run in a loop (indefinitely) simulating interacting with the nameservice app using all message types

IvantheTricourne:
Closes #168.
Closes #157.


IvantheTricourne:
Ran this for a bit this morning, and it seems that it might need more uniqueness:

```
interact: CallException JSON-RPC error -32603: Internal error. Data: Just (String "Error on broadcastTxCommit: Tx already exists in cache")
```

Happened after ~8000 transactions.

IvantheTricourne:
Gonna block this until #183 is addressed.

IvantheTricourne:
- Fixed an off-by-one error in Token transferring: previously, an account balance could never be depleted. That is, trying to buy something worth X with only X balance was impossible and triggered an `Insufficient Funds` exception from the Token module.
- Script now generates 2 random users for each thread

martyall (approved):

Issue #155, opened 2019-12-18
figure out what happened to the histograms in datadog


martyall:
"The buckets are present in prometheus, but are not being scraped by datadog" -Carl

fixes #155

PR #186, opened 2020-01-10
Make datadog work

### LINKED ISSUE #180: Move metrics to central location to be used by middleware and app
We want to remove the dependence on the datadog katip cribe and use prometheus for everything. This means that we'll need to move metrics stuff from the SDK into a common place

martyall:
fixes #180 #155 
Screen Shot 2020-01-10 at 9 50 48 AM
Screen Shot 2020-01-10 at 9 50 55 AM

### DIFF
--- a/hs-abci-examples/nameservice/docker-compose-test.yaml
+++ b/hs-abci-examples/nameservice/docker-compose-test.yaml
@@ -17,39 +17,28 @@ services:
     ports:
       - "26656-26657:26656-26657"
   nameservice:
-    entrypoint: /usr/local/bin/nameservice
     build:
       context: ../../.
       dockerfile: Dockerfile
     image: hs-abci:test
     environment:
-     - ES_HOST=elk
-     - ES_PORT=9200
-     - STATS_PORT=5555
+     - STATS_PORT=9200
     restart: always
-    depends_on:
-      - elk
+    entrypoint: /usr/local/bin/nameservice
     expose:
       - "26658"
-      - "5555:5555"
+      - "9200"
   datadog:
     image: datadog/agent:latest
-    depends_on:
-      - nameservice
-    restart: always
     environment:
      - DD_API_KEY=${DD_API_KEY}
     volumes:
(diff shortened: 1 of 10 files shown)


hs-abci-examples/nameservice/docker-compose-test.yaml
```diff
@@ -17,40 +17,28 @@ services:
...
-  elk:
-    image: sebp/elk:683
-    ports:
-      - "5601:5601"
-      - "9200:9200"
-      - "5044:5044"
-    expose:
-      - "9200"
-      - "5601"
+      com.datadoghq.ad.instances: '["{\"prometheus_url\":\"http://%%host%%:9200/metrics \",\"namespace\":\"nameservice\",\"metrics\":[\"count_buy\" \"count_set\" \"count_delete\" \"histogram_buy*\" \"histogram_set*\" \"histogram_delete*\"]}"]'
```
iostat: How come you're cherry-picking metrics here, but in the other docker-compose with the YAML config you just `metrics: ["*"]`?

hs-abci-sdk/src/Tendermint/SDK/BaseApp/Metrics/Prometheus.hs
```diff
@@ -205,7 +205,7 @@ evalMetrics state@MetricsState{..} = do
       start <- liftIO $ getCurrentTime
       a <- runT action
       end <- liftIO $ getCurrentTime
-      let time = fromRational . (* 1000.0) . toRational $ (end `diffUTCTime` start)
+      let time = realToFrac (end `diffUTCTime` start)
```
iostat, line 208: suggestion: use `ms` as your unit -- by the looks of your graphs your operations are in that time scale and you're just giving yourself floating imprecision for no good reason by choosing seconds

hs-abci-examples/nameservice/docker-compose-test.yaml
martyall: this actually shouldn't even be here, I wanted to remove all logging and metrics from the test compose file

PR #187, opened 2020-01-10
Add Nonce

### LINKED ISSUE #183: Add nonce to transactions
We want to be able to commit multiple semantically equivalent transactions (e.g. a transfer between the same 2 accounts for the same value). Currently, this fails during the `checkTx` phase with the error: 

```
"Error on broadcastTxCommit: Tx already exists in cache"
```

To address this, we should propagate the nonce from the [Auth module](https://github.com/f-o-a-m/hs-abci/blob/docker-stack-cache/hs-abci-sdk/src/Tendermint/SDK/Modules/Auth/Types.hs#L43) to transactions performed by respective account. This will probably take the shape of an _AnteHandler_ check which happens before message routing and validation. This AnteHandler is responsible for validating that the appropriate nonce is updated for a given transaction.

IvantheTricourne:
Closes #183


IvantheTricourne:
- [x] Nonce checking 
- [x] Nonce querying

But, for some reason, the `Tx` nonce is always 0.

EDIT: Fixed. Nonce updating should only happen during `DeliverTx`. Checking (i.e., throwing exceptions from checking the nonce) only happens during `CheckTx`

martyall (approved):

PR #194, opened 2020-01-16
Query params

martyall:
Allows for query params (from servant) to be part of query URL

### DIFF
--- a/hs-abci-sdk/test/Tendermint/SDK/Test/QuerySpec.hs
+++ b/hs-abci-sdk/test/Tendermint/SDK/Test/QuerySpec.hs
@@ -1,0 +1,54 @@
+module Tendermint.SDK.Test.QuerySpec (spec) where
+
+import           Control.Lens                         ((^.))
+import qualified Data.ByteArray.Base64String          as Base64
+import           Data.Text                            (Text)
+import qualified Network.ABCI.Types.Messages.Request  as Req
+import qualified Network.ABCI.Types.Messages.Response as Resp
+import qualified Tendermint.SDK.Application           as App
+import qualified Tendermint.SDK.Application.Module    as M
+import qualified Tendermint.SDK.BaseApp               as BA
+import qualified Tendermint.SDK.BaseApp.Logger.Katip  as KL
+import qualified Tendermint.SDK.BaseApp.Transaction   as T
+import           Tendermint.SDK.Codec                 (HasCodec (..))
+import qualified Tendermint.SDK.Test.SimpleStorage    as SS
+import           Tendermint.SDK.Types.Message         (Msg (..))
+import           Tendermint.SDK.Types.Transaction     (PreRoutedTx (..),
+                                                       Tx (..))
+import           Test.Hspec
+
+type Effs = SS.SimpleStorage ': BA.BaseApp BA.CoreEffs
+
+spec :: Spec
+spec = beforeAll (BA.makeContext (KL.InitialLogNamespace "test" "spec") Nothing) $
+  describe "Query tests" $ do
+    let modules :: App.Modules '[SS.SimpleStorageM Effs] Effs
+        modules = SS.simpleStorageModule App.:+ App.NilModules
+        ssServer = M.queryRouter modules
+        handler = App.moduleTxDeliverer SS.simpleStorageModule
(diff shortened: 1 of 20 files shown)


hs-abci-test-utils/src/Tendermint/Utils/Client.hs
```diff
@@ -36,40 +39,75 @@ instance RunClient (ReaderT RPC.Config IO) where
...
   , clientResponseRaw  :: Resp.Query
   }
 
+addQueryParamsToPath
+  :: QueryStringList
+  -> Text
+  -> Text
+addQueryParamsToPath qs path =
+  let qParams = intercalate "&" $ map (\(n,v) -> n <> "=" <> v) qs
+  in path <> "?" <> qParams
```
IvantheTricourne: This seems to be adding `?` even though there aren't any `qs`.

IvantheTricourne: For example: https://travis-ci.com/f-o-a-m/hs-abci/jobs/276480230#L1304

hs-abci-sdk/test/Tendermint/SDK/Test/SimpleStorage.hs
```diff
@@ -0,0 +1,182 @@
...
+          updateCount (Count updateCountTxCount)
+
+--------------------------------------------------------------------------------
+-- Server
+--------------------------------------------------------------------------------
+
+type CountStoreContents = '[(CountKey, Count)]
+
+type GetMultipliedCount =
+     "multiply"
```
IvantheTricourne: Probably mean `multiplier`, as indicated by the tests: https://github.com/f-o-a-m/hs-abci/pull/194/files#diff-f3e7ffe5a8bd3efb979298b272d06c93R46

IvantheTricourne:
So it seems that the full path to the query isn't being properly populated when it gets here: https://github.com/f-o-a-m/hs-abci/pull/194/files?file-filters%5B%5D=.hs#diff-18418e639e179d33970efb24cc9030f1R63

I snuck in a log to see what the path was that was triggering the `Path not found.` exception, and it looks like this:

```
queryPath = "multiplied?factor=4"
```

hs-abci-sdk/test/Tendermint/SDK/Test/QuerySpec.hs
```diff
@@ -0,0 +1,54 @@
...
+              , txGas = 0
+              , txSignature = undefined
+              , txSignBytes = undefined
+              , txSigner = undefined
+              , txNonce = undefined
+              }
+        txContext <- T.newTransactionContext tx
+        _ <- SS.evalToIO ctx . T.eval txContext $ handler tx
+        let q = Req.Query
+              -- TODO -- this shouldn't require / count
```
IvantheTricourne, line 45: I'm guessing that this isn't the case anymore?

hs-abci-sdk/test/Tendermint/SDK/Test/SimpleStorage.hs
```diff
@@ -0,0 +1,184 @@
...
+-- Server
+--------------------------------------------------------------------------------
+
+type CountStoreContents = '[(CountKey, Count)]
+
+type GetMultipliedCount =
+     "manipulated"
+  :> Capture "subtract" Integer
+  :> QueryParam' '[Required, Strict] "factor" Integer
+  :> BaseApp.Leaf Count
```
IvantheTricourne, line 128: Question: Aside from `Leaf`, how is this different from how servant normally does things?

IvantheTricourne (approved):

hs-abci-sdk/test/Tendermint/SDK/Test/QuerySpec.hs
martyall: yeah but I'm gonna merge it anyway

hs-abci-sdk/test/Tendermint/SDK/Test/SimpleStorage.hs
martyall: Servant has the option for `GET, PUT, POST` etc, we don't have any of these but we still need to specify the return type

PR #200, opened 2020-01-23
Module router refactor

martyall:


hs-abci-examples/simple-storage/test/SimpleStorage/Test/E2ESpec.hs
```diff
@@ -1,63 +1,121 @@
...
+
+  in ClientConfig
+       { clientGetNonce = getNonce
+       , clientRPC = rpcConfig
+       }
+
+type TxClientM = ReaderT ClientConfig IO
+
+runTxClientM :: TxClientM a -> IO a
+runTxClientM m = runReaderT m txClientConfig
```
IvantheTricourne, line 104: Any reason why this doesn't live in the utils lib? It's also in the nameservice tests.

martyall: because it requires the api for the auth module and currently I didn't implement a way to do that without generating an api for an entire application, not just a single module. I will go back later to do this, not this PR

PR #201, opened 2020-01-28
Impl bank module

### LINKED ISSUE #189: Move nameservice token module to SDK
The nameservice token module (i.e., actions to transfer, burn and mint tokens) should be moved to the SDK. From there, it should be updated to use account balances with proper coin denominations (e.g. a  nameservice coin) via the SDK's auth module.

The FaucetAccount action should remain in nameservice.

IvantheTricourne:
Closes #189 

Includes stricter nonce checking logic: `TxNonce` should be +1 of `accountNonce`


hs-abci-examples/nameservice/src/Nameservice/Modules/Nameservice/Keeper.hs
```diff
@@ -84,7 +104,7 @@ deleteName DeleteName{..} = do
       if whoisOwner /= deleteNameOwner
         then throw $ InvalidDelete "Deleter must be the owner."
         else do
-          mint deleteNameOwner whoisPrice
+          mint deleteNameOwner (Coin "nameservice" whoisPrice)
```
martyall: we should set a constant 

```haskell
nameserviceCoinId :: CoinId
nameserviceCoinId = "nameservice"
```

to avoid spelling error

hs-abci-sdk/src/Tendermint/SDK/Application/AnteHandler.hs
```diff
@@ -31,32 +31,49 @@ applyAnteHandler
...
-        unless (accountNonce <= txNonce) $
-          throwSDKError (NonceException accountNonce txNonce)
-        pure a
+    preMAcnt <- A.getAccount msgAuthor
+    case preMAcnt of
+      Just A.Account{accountNonce} -> do
+        let expectedNonce = accountNonce + 1
+        unless (txNonce == expectedNonce) $
+          throwSDKError (NonceException expectedNonce txNonce)
       Nothing -> do
```
martyall: since this is no longer possible when using the `createAccountAnteH`, this should just throw an error.

hs-abci-sdk/src/Tendermint/SDK/Application/AnteHandler.hs
```diff
@@ -31,32 +31,49 @@ applyAnteHandler
...
     result <- txApplication tx
-    A.putAccount msgAuthor $
-      account { A.accountNonce = A.accountNonce account + 1}
+    postMAcnt <- A.getAccount msgAuthor
+    case postMAcnt of
+      Just acnt@A.Account{accountNonce} -> do
+        A.putAccount msgAuthor $
+          acnt { A.accountNonce = accountNonce + 1}
+      -- @NOTE: no-op when no nonce is availble to update
+      Nothing -> pure ()
```
martyall, line 54: similarly here it's not really possible, although maybe in the future we support deleting accounts so it's harmless as is.

hs-abci-sdk/src/Tendermint/SDK/Application/AnteHandler.hs
```diff
@@ -31,32 +31,49 @@ applyAnteHandler
...
     pure result
 
 baseAppAnteHandler
   :: Members A.AuthEffs r
   => Member (Error AppError) r
   => AnteHandler r
 baseAppAnteHandler = mconcat $
-  [ nonceAnteHandler
```
martyall, line 46: maybe make a comment about how these are run in order starting from the first in the list

hs-abci-sdk/src/Tendermint/SDK/Modules/Auth/Types.hs
```diff
@@ -5,27 +5,79 @@ module Tendermint.SDK.Modules.Auth.Types where
...
-import           Tendermint.SDK.Codec         (HasCodec (..))
+import           Tendermint.SDK.Codec         (HasCodec (..),
+                                               defaultSDKAesonOptions)
 import           Tendermint.SDK.Types.Address (Address)
+import           Web.HttpApiData              (FromHttpApiData (..),
+                                               ToHttpApiData (..))
 
 type AuthModule = "auth"
 
+newtype CoinId = CoinId { unCoinId :: Text } deriving (Eq, Show, Generic)
```
martyall, line 55: I'm starting to lean toward the fact we shouldn't really use this proto-3 suite in the sdk itself, but rather reserve it as a convenience for developing apps more quickly. This is why I've included proto files for things like `Account` and `RawTransaction` in the `/protos` dir of the sdk.

It adds more boilerplate, but i think its ultimately cleaner given that protolens is a more robust library and at some point we might be taking protofiles from a central repository for cosmos modules

hs-abci-sdk/src/Tendermint/SDK/Modules/Bank/Keeper.hs
```diff
@@ -0,0 +1,108 @@
...
+  let zeroBalance = Auth.Coin cid 0
+  case mAcnt of
+    Nothing -> pure zeroBalance
+    Just (Auth.Account coins _) ->
+      let mCoin = find (\(Auth.Coin cid1 _) -> cid == cid1) coins
+      in pure $ fromMaybe zeroBalance mCoin
+
+replaceCoinValue :: Auth.Coin -> [Auth.Coin] -> [Auth.Coin]
+replaceCoinValue c [] = [c]
+replaceCoinValue c@(Auth.Coin cid _) (c1@(Auth.Coin cid1 _):rest) =
```
martyall: more common to use `cid'`

hs-abci-sdk/src/Tendermint/SDK/Modules/Bank/Keeper.hs
```diff
@@ -0,0 +1,108 @@
...
+  => Members Auth.AuthEffs r
+  => Members BankEffs r
+  => Address
+  -> Auth.Coin
+  -> Address
+  -> Sem r ()
+transfer addr1 (Auth.Coin cid amount) addr2 = do
+  -- check if addr1 has amt
+  (Auth.Coin _ addr1Bal) <- getCoinBalance addr1 cid
+  if addr1Bal > amount
```
martyall: should be `>=` no?

hs-abci-sdk/src/Tendermint/SDK/Modules/Bank/Query.hs
```diff
@@ -0,0 +1,43 @@
...
+import           Tendermint.SDK.Modules.Bank.Keeper (getCoinBalance)
+import           Tendermint.SDK.Types.Address       (Address)
+
+--------------------------------------------------------------------------------
+-- | Query Api
+--------------------------------------------------------------------------------
+
+type GetAddressCoinBalance =
+     "balance"
+  :> QueryParam' '[Required, Strict] "address" Address
```
martyall: let's go ahead and use `QA Address` here since someone might be querying for historical balances. We can't actually implement that at the moment, but we can go ahead and use the type

hs-abci-sdk/src/Tendermint/SDK/Modules/Bank/Query.hs
```diff
@@ -0,0 +1,43 @@
...
+  :> QueryParam' '[Required, Strict] "coin_id" Auth.CoinId
+  :> BaseApp.Leaf Auth.Coin
+
+getAddressCoinBalance
+  :: Members Auth.AuthEffs r
+  => Address
```
martyall: This would change to `QueryArgs Address`

hs-abci-sdk/src/Tendermint/SDK/Modules/Bank/Types.hs
```diff
@@ -0,0 +1,89 @@
+{-# OPTIONS_GHC -fno-warn-orphans #-}
```
martyall: there's no need for orphan instances here, move any instance declarations to the module where `Address` is defined

hs-abci-sdk/src/Tendermint/SDK/Modules/Bank/Types.hs
```diff
@@ -0,0 +1,89 @@
...
+  | PutOnNonExistentAccount Address
+
+instance BaseApp.IsAppError BankError where
+  makeAppError (InsufficientFunds msg) =
+    BaseApp.AppError
+    { appErrorCode = 1
+    , appErrorCodespace = "bank"
+    , appErrorMessage = msg
+    }
+  makeAppError (PutOnNonExistentAccount addr) =
```
martyall: This error shouldn't be here. i know we've talked about this a few times, but this is one instance where it actually makes sense to create an account and put funds into it.

hs-abci-sdk/src/Tendermint/SDK/Modules/Bank/Keeper.hs
```diff
@@ -0,0 +1,108 @@
...
+putCoinBalance
+  :: Members Auth.AuthEffs r
+  => Members BankEffs r
+  => Address
+  -> Auth.Coin
+  -> Sem r ()
+putCoinBalance address coin = do
+  mAcnt <- Auth.getAccount address
+  acnt <- case mAcnt of
+            Nothing -> throw (PutOnNonExistentAccount address)
```
martyall: should create account (see note below)

hs-abci-sdk/src/Tendermint/SDK/Modules/Bank/Types.hs
```diff
@@ -0,0 +1,89 @@
...
+instance FromHttpApiData Address where
+  parseQueryParam = Right . Address . fromString . unpack
+
+--------------------------------------------------------------------------------
+-- Exceptions
+--------------------------------------------------------------------------------
+
+data BankError =
+    InsufficientFunds Text
+  | PutOnNonExistentAccount Address
```
martyall: shoudn't be here (see below)

hs-abci-sdk/src/Tendermint/SDK/Modules/TypedMessage.hs
```diff
@@ -0,0 +1,25 @@
+module Tendermint.SDK.Modules.TypedMessage where
```
martyall: this module should be removed, we don't use it anymore

martyall (changes requested):

hs-abci-sdk/src/Tendermint/SDK/Modules/Bank/Keeper.hs
IvantheTricourne: oops... don't know how this snuck back in there.

hs-abci-sdk/src/Tendermint/SDK/Application/AnteHandler.hs
IvantheTricourne: I actually tested this out yesterday... this `Nothing` case is always triggered by `checkTx` , so I ended up leaving this alone.

hs-abci-sdk/src/Tendermint/SDK/Modules/Auth/Types.hs
IvantheTricourne: Should I change this? Or perhaps make an issue after this to purge it from the SDK?

martyall: We should just do it now, it's not that bad

martyall (approved):

PR #205, opened 2020-02-03
subscribe

charlescrain:


charlescrain:
We mentioned making it a conduit source but I had trouble imagining a clean way to do this. For now WS methods take in a handler.

@blinky3713 Let's discuss how to wrap it as a conduit source when you get a chance

martyall (approved):

PR #227, opened 2020-03-10
Hide literate haskell main from github markdown

alexfmpe:
It's possible to hide the main from the rendered version (assuming this is desired).


martyall:
oh cool, I didn't know you could do that

PR #245, opened 2020-08-15
Fix github fields

alexfmpe:


hs-abci-extra/package.yaml
```diff
@@ -1,6 +1,6 @@
 name:                hs-abci-extra
 version:             0.1.0.0
-github:              "https://github.com/f-o-a-m/kepler/hs-abci-server"
```
alexfmpe, line 3: The http format trips up hpack in newer versions

alexfmpe:
Relatedly, the nameservice/simple-storage apps' github field also trip up recent hpack, but I think it's a bug in the latter: https://github.com/sol/hpack/pull/399

martyall (approved):

Issue #252, opened 2020-12-17
Module for Validator Set Managment


UnitylChaos:
With #247 merged, Application developers can update the validator set through the EndBlock handler. Given that any system for dynamic validator sets will have some common components, it may be worth making a module for these in the SDK.

This is my basic draft for what I think would be useful Validator module:

* Store
  * CurrentValidators :: Map Pubkey Natural
  * QueuedUpdates :: Map Pubkey Natural
* Keeper
  * GetValidators :: Map Pubkey Natural
  * GetPowerOf :: Pubkey -> Natural
  * GetQueuedUpdates :: Map Pubkey Natural
  * GetQueuedUpdate :: Pubkey -> Maybe Natural
    * Maybe because update to 0 means the validator is being removed, not that there is no update
  * QueueUpdate :: Pubkey -> Natural -> ()
  * QueueUpdates :: [(Pubkey, Natural)] -> ()
* Query
  * GetValidators
  * GetPowerOf
  * GetQueuedUpdates
  * GetQueuedUpdate
* EndBlock = do
  * Update CurrentValidators with QueuedUpdates
  * Construct EndBlockResult with QueuedUpdates to send to tendermint
  * Reset QueuedUpdates for next block

This could be optionally extended to store historical changes too, by replacing QueuedUpdates with a [Map Pubkey Natural], indexed by block height, but that might be a lot of state space for chains with lots of changes, and almost certainly wouldn't be needed by all chains. (Likewise for historical validator sets)

martyall:
this is something that I know (embarrassingly) little about so I would follow your lead on the "what is it supposed to do" part and reserve my comments for the code itself. 

Is this module that you're describing responsible for updating the state needed for block proposal and tendermint consensus to work? Ive never really understood how our example applications work since we run a tendermint docker image out of the box with no modifications...

martyall:
it might be worth proposing in this issue the concrete message and storage types though

UnitylChaos:
This module would basically do two things:
1. Act as a queue for other modules which need to make changes to the Tendermint validator set. For example a staking module.
2. Keep a copy of the current Tendermint validator set up to date, for modules which might need that information. This is done since (AFAIK) there is no good way to query the validator set from Tendermint, so the only way for a module to know who the validators are is to follow all the updates.

[...]

PR #253, opened 2020-12-20
Validators module

UnitylChaos:
Initial draft of a module for apps to manage validator sets (#252).

Changes from what was described in that issue:
- Addition of store Var to keep a set of the current validator pubkeys, since store Map doesn't have a method to get all keys/values
- Used List type for QueuedUpdates, so that adding an update would be fast, and Map like behavior can be constructed from a foldl
- Used Word64 instead of Natural. Still non-negative, but bounded since Tendermint has a 64bit bound on validator power anyway.
- Didn't implement Queries for QueuedUpdates, since it is always reset in EndBlock, and query is for committed state (it would always be [])

I'm not very familiar with the Query system, so that is in particular need of review.
Also Weeder is going to fail until this has some tests, and I imagine that is going to take a while because I'm not familiar with the Kepler testing system, and proper testing will require multiple Tendermint docker instances.

### DIFF
--- a/hs-abci-sdk/src/Tendermint/SDK/Modules/Validators/EndBlock.hs
+++ b/hs-abci-sdk/src/Tendermint/SDK/Modules/Validators/EndBlock.hs
@@ -49,4 +49,4 @@ endBlock _ = do
   pure $ EndBlockResult (map convertToValUp (Map.assocs updatesMap)) Nothing
   where
     convertToValUp (PubKey_ key, power) =
-      ABCI.ValidatorUpdate key (ABCI.WrappedVal (fromIntegral power))
+      ABCI.ValidatorUpdate (Just key) (ABCI.WrappedVal (fromIntegral power))
--- a/hs-abci-sdk/src/Tendermint/SDK/Modules/Validators/EndBlock.hs
+++ b/hs-abci-sdk/src/Tendermint/SDK/Modules/Validators/EndBlock.hs
@@ -1,0 +1,50 @@
+module Tendermint.SDK.Modules.Validators.EndBlock where
+
+import           Control.Monad                            (foldM)
+import qualified Data.Map.Strict                          as Map
+import qualified Data.Set                                 as Set
+import qualified Network.ABCI.Types.Messages.FieldTypes   as ABCI
+import qualified Network.ABCI.Types.Messages.Request      as Request
+import           Polysemy                                 (Members, Sem)
+import           Tendermint.SDK.BaseApp                   (BlockEffs,
+                                                           EndBlockResult (..))
+import qualified Tendermint.SDK.BaseApp.Store.List        as L
+import qualified Tendermint.SDK.BaseApp.Store.Map         as M
+import qualified Tendermint.SDK.BaseApp.Store.Var         as V
+import           Tendermint.SDK.Modules.Validators.Keeper
+import           Tendermint.SDK.Modules.Validators.Store
+import           Tendermint.SDK.Modules.Validators.Types
+
+
+endBlock
+  :: Members BlockEffs r
+  => Members ValidatorsEffs r
+  => Request.EndBlock
+  -> Sem r EndBlockResult
+endBlock _ = do
+  updatesMap <- getQueuedUpdates
+  curValKeySet <- getValidatorsKeys
+
+  -- update the Validators map and key set
+  newValKeySet <- foldM (\cvks (key, newPower) ->
+      if newPower == 0 then do
+        -- delete from Validators map and key set
+        M.delete key validatorsMap
+        return (Set.delete key cvks)
+      else do
+        -- update power in Validators map and ensure key is in key set
+        M.insert key newPower validatorsMap
+        return (Set.insert key cvks)
+    ) curValKeySet (Map.assocs updatesMap)
+
+  -- store new set of validator keys
+  V.putVar (KeySet newValKeySet) validatorsKeySet
+
+  -- reset the updatesList to empty
+  L.deleteWhen (const True) updatesList
+
+  -- return EndBlockResult with validator updates for Tendermint
+  pure $ EndBlockResult (map convertToValUp (Map.assocs updatesMap)) Nothing
+  where
+    convertToValUp (PubKey_ key, power) =
+      ABCI.ValidatorUpdate (Just key) (ABCI.WrappedVal (fromIntegral power))
--- a/hs-abci-sdk/src/Tendermint/SDK/Modules/Validators/EndBlock.hs
+++ b/hs-abci-sdk/src/Tendermint/SDK/Modules/Validators/EndBlock.hs
@@ -1,6 +1,8 @@
 module Tendermint.SDK.Modules.Validators.EndBlock where
 
-import           Control.Monad                            (foldM)
+import           Control.Monad.State                      (MonadTrans (lift),
+                                                           execStateT, forM_,
+                                                           modify)
 import qualified Data.Map.Strict                          as Map
 import qualified Data.Set                                 as Set
 import qualified Network.ABCI.Types.Messages.FieldTypes   as ABCI
@@ -26,16 +28,16 @@ endBlock _ = do
   curValKeySet <- getValidatorsKeys
 
   -- update the Validators map and key set
-  newValKeySet <- foldM (\cvks (key, newPower) ->
+  newValKeySet <- flip execStateT curValKeySet $
+    forM_ (Map.toList updatesMap) $ \(key, newPower) ->
       if newPower == 0 then do
         -- delete from Validators map and key set
-        M.delete key validatorsMap
-        return (Set.delete key cvks)
+        lift $ M.delete key validatorsMap
+        modify $ Set.delete key
       else do
         -- update power in Validators map and ensure key is in key set
-        M.insert key newPower validatorsMap
-        return (Set.insert key cvks)
-    ) curValKeySet (Map.assocs updatesMap)
+        lift $ M.insert key newPower validatorsMap
+        modify $ Set.insert key
 
   -- store new set of validator keys
   V.putVar (KeySet newValKeySet) validatorsKeySet
@@ -47,4 +49,4 @@ endBlock _ = do
   pure $ EndBlockResult (map convertToValUp (Map.assocs updatesMap)) Nothing
   where
     convertToValUp (PubKey_ key, power) =
-      ABCI.ValidatorUpdate (Just key) (ABCI.WrappedVal (fromIntegral power))
+      ABCI.ValidatorUpdate key (ABCI.WrappedVal (fromIntegral power))
--- a/hs-abci-sdk/src/Tendermint/SDK/Modules/Validators/Keeper.hs
+++ b/hs-abci-sdk/src/Tendermint/SDK/Modules/Validators/Keeper.hs
@@ -6,6 +6,7 @@ import qualified Data.Map.Strict                         as Map
 import           Data.Maybe                              (fromMaybe)
 import qualified Data.Set                                as Set
 import           Data.Word                               (Word64)
+import           Network.ABCI.Types.Messages.FieldTypes
 import           Polysemy                                (Members, Sem,
                                                           interpret, makeSem)
 import           Polysemy.Error                          (Error)
@@ -19,13 +20,10 @@ import           Tendermint.SDK.Modules.Validators.Types
 
 
 data ValidatorsKeeper m a where
-  GetValidators :: ValidatorsKeeper m (Map.Map PubKey_ Word64)
   GetValidatorsKeys :: ValidatorsKeeper m (Set.Set PubKey_)
   GetPowerOf :: PubKey_ -> ValidatorsKeeper m Word64
   GetQueuedUpdates :: ValidatorsKeeper m (Map.Map PubKey_ Word64)
-  GetQueuedUpdate :: PubKey_ -> ValidatorsKeeper m (Maybe Word64)
   QueueUpdate :: PubKey_ -> Word64 -> ValidatorsKeeper m ()
-  QueueUpdates :: [(PubKey_, Word64)] -> ValidatorsKeeper m ()
 
 makeSem ''ValidatorsKeeper
 
@@ -36,21 +34,11 @@ eval
   => Sem (ValidatorsKeeper : r) a
   -> Sem r a
 eval = interpret (\case
-  GetValidators -> getValidatorsF
   GetValidatorsKeys -> getValidatorsKeysF
   GetPowerOf key -> getPowerOfF key
   GetQueuedUpdates -> getQueuedUpdatesF
-  GetQueuedUpdate key -> getQueuedUpdateF key
   QueueUpdate key power -> queueUpdateF key power
-  QueueUpdates updates -> queueUpdatesF updates
   )
-
-getValidatorsF
-  :: Members [ReadStore, Error AppError] r
-  => Sem r (Map.Map PubKey_ Word64)
-getValidatorsF = do
-  keyList <- fmap Set.toList getValidatorsKeysF
-  fmap Map.fromList $ mapM (\k -> fmap (\p -> (k, p)) (getPowerOfF k)) keyList
 
 getValidatorsKeysF
   :: Members [ReadStore, Error AppError] r
@@ -68,27 +56,18 @@ getPowerOfF key =
 getQueuedUpdatesF
   :: Members [ReadStore, Error AppError] r
   => Sem r (Map.Map PubKey_ Word64)
-getQueuedUpdatesF = L.foldl (\m ValidatorUpdate{..} ->
-  Map.alter (Just . fromMaybe power) key m) Map.empty updatesList
-
-getQueuedUpdateF
-  :: Members [ReadStore, Error AppError] r
-  => PubKey_
-  -> Sem r (Maybe Word64)
-getQueuedUpdateF key =
-  fmap (Map.lookup key) getQueuedUpdatesF
+getQueuedUpdatesF = L.foldl (\m (ValidatorUpdate_ ValidatorUpdate{..}) ->
+  Map.alter (Just . fromMaybe (toWord validatorUpdatePower)) (PubKey_ validatorUpdatePubKey) m) Map.empty updatesList
+  where
+    toWord (WrappedVal x) = fromIntegral x
 
 queueUpdateF
   :: Members [ReadStore, WriteStore, Error AppError] r
   => PubKey_
   -> Word64
   -> Sem r ()
-queueUpdateF key power =
-  L.append (ValidatorUpdate key power) updatesList
-
-queueUpdatesF
-  :: Members [ReadStore, WriteStore, Error AppError] r
-  => [(PubKey_, Word64)]
-  -> Sem r ()
-queueUpdatesF = mapM_ $ uncurry queueUpdateF
+queueUpdateF (PubKey_ key) power =
+  L.append (ValidatorUpdate_(ValidatorUpdate key (wrapInt power))) updatesList
+  where
+    wrapInt p = WrappedVal (fromIntegral p)
 
--- a/hs-abci-sdk/src/Tendermint/SDK/Modules/Validators/Keeper.hs
+++ b/hs-abci-sdk/src/Tendermint/SDK/Modules/Validators/Keeper.hs
@@ -3,7 +3,7 @@
 module Tendermint.SDK.Modules.Validators.Keeper where
 
 import qualified Data.Map.Strict                         as Map
-import           Data.Maybe                              (fromMaybe)
+import           Data.Maybe                              (fromJust, fromMaybe)
 import qualified Data.Set                                as Set
 import           Data.Word                               (Word64)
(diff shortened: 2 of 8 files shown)


hs-abci-sdk/src/Tendermint/SDK/Modules/Validators/Types.hs
```diff
@@ -0,0 +1,55 @@
...
+
+
+data ValidatorUpdate = ValidatorUpdate
+    { key   :: PubKey_
+    , power :: Word64
+    }
+    deriving Generic
+instance A.ToJSON ValidatorUpdate
+instance A.FromJSON ValidatorUpdate
+instance HasCodec ValidatorUpdate where
```
blinky3713, line 38: For the state objects we would like to use a `.proto` file to describe the objects, see for example how the `Auth` module is done. https://github.com/f-o-a-m/kepler/tree/master/hs-abci-sdk/protos/modules

There are several reasons why we do this, you can see this discussion if you're interested: https://github.com/cosmos/cosmos-rust/issues/14#issuecomment-748174290

hs-abci-sdk/src/Tendermint/SDK/Modules/Validators/Keeper.hs
```diff
@@ -0,0 +1,94 @@
...
+
+
+data ValidatorsKeeper m a where
+  GetValidators :: ValidatorsKeeper m (Map.Map PubKey_ Word64)
+  GetValidatorsKeys :: ValidatorsKeeper m (Set.Set PubKey_)
+  GetPowerOf :: PubKey_ -> ValidatorsKeeper m Word64
+  GetQueuedUpdates :: ValidatorsKeeper m (Map.Map PubKey_ Word64)
+  GetQueuedUpdate :: PubKey_ -> ValidatorsKeeper m (Maybe Word64)
+  QueueUpdate :: PubKey_ -> Word64 -> ValidatorsKeeper m ()
+  QueueUpdates :: [(PubKey_, Word64)] -> ValidatorsKeeper m ()
```
martyall: You don't actually need `QueueUpdates` as a summand since it's implemented in terms of folding `QueueUpdate`, you can just write a function 

`queueUpdates :: Members ValidatorsEffs r => [(PubKey_, Word64)] -> Sem r ()` 

directly.

hs-abci-sdk/src/Tendermint/SDK/Modules/Validators/Keeper.hs
```diff
@@ -0,0 +1,94 @@
...
+                                                          WriteStore)
+import qualified Tendermint.SDK.BaseApp.Store.List       as L
+import qualified Tendermint.SDK.BaseApp.Store.Map        as M
+import qualified Tendermint.SDK.BaseApp.Store.Var        as V
+import           Tendermint.SDK.Modules.Validators.Store
+import           Tendermint.SDK.Modules.Validators.Types
+
+
+data ValidatorsKeeper m a where
+  GetValidators :: ValidatorsKeeper m (Map.Map PubKey_ Word64)
```
martyall: You probably don't need a `GetValidators` summand because you can see below that it's implemented in terms of `GetValidatorsKeys` and `GetPowerOf`

hs-abci-sdk/src/Tendermint/SDK/Modules/Validators/Keeper.hs
```diff
@@ -0,0 +1,94 @@
...
+  GetValidatorsKeys :: ValidatorsKeeper m (Set.Set PubKey_)
+  GetPowerOf :: PubKey_ -> ValidatorsKeeper m Word64
+  GetQueuedUpdates :: ValidatorsKeeper m (Map.Map PubKey_ Word64)
+  GetQueuedUpdate :: PubKey_ -> ValidatorsKeeper m (Maybe Word64)
```
martyall: You don't need `GetQueuedUpdate` as a summand since it's implement in terms of `GetQueuedUpdates`

hs-abci-sdk/src/Tendermint/SDK/Modules/Validators/EndBlock.hs
```diff
@@ -0,0 +1,50 @@
...
+  :: Members BlockEffs r
+  => Members ValidatorsEffs r
+  => Request.EndBlock
+  -> Sem r EndBlockResult
+endBlock _ = do
+  updatesMap <- getQueuedUpdates
+  curValKeySet <- getValidatorsKeys
+
+  -- update the Validators map and key set
+  newValKeySet <- foldM (\cvks (key, newPower) ->
```
martyall: I think it would be better to implement this with a normal fold in `StateT (Set PubKey_) (Sem r) ()` rather than `foldM` because in the end you are doing a lot of extra work concatenating all of those sets at the end rather than single targeted updates.

hs-abci-sdk/src/Tendermint/SDK/Modules/Validators/EndBlock.hs
```diff
@@ -0,0 +1,50 @@
...
+      if newPower == 0 then do
+        -- delete from Validators map and key set
+        M.delete key validatorsMap
+        return (Set.delete key cvks)
+      else do
+        -- update power in Validators map and ensure key is in key set
+        M.insert key newPower validatorsMap
+        return (Set.insert key cvks)
+    ) curValKeySet (Map.assocs updatesMap)
```
martyall: why use `assocs` and not `toList` ?

martyall (changes requested):
Overall looks good! Let's go ahead and get these changes in and then I will do a second pass.

Two kind of interesting things are happening here that haven't come up before:

1. You are using a `Set` to keep track of the keys that belong to a certain store map.
2. There is a state object (`ValidatorUpdate`) that is not accessible to the outside world.

In both cases you opted to use JSON to implement the `HasCodec` instance. In general I am not a fan of this because it makes it more difficult to operate on these values across languages and applications. There are a few reasons why, you can see this discussion if you're interested: https://github.com/cosmos/cosmos-rust/issues/14#issuecomment-748174290

I'm not even sure I care about addressing this now, but:
(2) above would easily solved by just implementing a protobuf definition, but I have a feeling (1) is going to come up over and over again because it does with solidity.

martyall:
regarding a proper tests setup with multiple docker images, I am not really worried about that in this PR but feel free to file an issue. It seems simple enough that things should just work 🤞

hs-abci-sdk/src/Tendermint/SDK/Modules/Validators/Keeper.hs
UnitylChaos: Fair point. Will remove.

hs-abci-sdk/src/Tendermint/SDK/Modules/Validators/Keeper.hs
UnitylChaos: :+1:

hs-abci-sdk/src/Tendermint/SDK/Modules/Validators/Keeper.hs
UnitylChaos: :+1:

hs-abci-sdk/src/Tendermint/SDK/Modules/Validators/EndBlock.hs
UnitylChaos: I'm not really following you here. I'm not very familiar with StateT, but i get the idea that it can be used in place of a foldM (sorta). If the only issue is not wanting to do the repeated Set.insert / Set.delete, would something like this be acceptable?

```
  (adds, rems) <- foldM (\(a,r) (key, newPower) ->
    if newPower == 0 then do
       M.delete key validatorsMap
       return (a, key : r) 
    else do
       M.insert key newPower validatorsMap
       return (key : a, r)
    ) ([], []) (Map.toList updatesMap)

  let newValKeySet = Set.union (curValKeySet Set.\\ Set.fromList rems) (Set.fromList adds)
```
Which would just accumulate the items to be added/removed through list building and then do two Set operations at the end.

hs-abci-sdk/src/Tendermint/SDK/Modules/Validators/EndBlock.hs
UnitylChaos: No real reason, I forgot the toList function and saw assocs first on the hackage page. Will replace.

UnitylChaos:
To your comment points:

I think the "Set of Map keys" pattern could be added as a feature in Map. Since IAVL has the IterateRange methods and the gRPC layer has the List method, it should be possible to get all the key/value pairs for a Map substore. I can write up an issue if you think this is worth pursuing. 


`ValidatorUpdate` is basically just a redefinition of the `ValidatorUpdate` data type from ABCI, but using `PubKey_` instead of `Maybe ABCI.PubKey` (Not clear to me why `ABCI.ValidatorUpdate` has a `Maybe` on it's `PubKey`) so that `HasCodec` could easily implemented on both (using Generic and JSON). I would generally prefer to just reuse the ABCI types directly, but we can't really put `HasCodec` and `RawKey` implementations in ABCI without creating a circular dependency mess between ABCI and the SDK.

I can probably do this more cleanly with newtype and by reusing the proto stuff from ABCI to implement `HasCodec` / `RawKey` in types.hs.

Also ValidatorUpdate is not externally accessible partly because it's only being used as a way to queue updates from different deliverTx into the endblock, it's never actually stored in committed state. Which is somewhat weird, but I can't really see a better way to do this other than creating some entirely new system for local block processing storage.

I agree with the larger point about cross platform compatibility and standardization, so I'll switch all of this over to proto.

hs-abci-sdk/src/Tendermint/SDK/Modules/Validators/EndBlock.hs
martyall: I find it kind of interesting that you can hack polysemy but aren't really starting with a knowledge of `mtl`, not everyone's haskell trajectory goes like that 👏 


with `StateT` it would look like 

```haskell
  newValKeySet <- flip execStateT curValKeySet $
    forM_ (Map.assocs updatesMap) $ (key, newPower) ->
      if newPower == 0 
        then do
          -- delete from Validators map and key set
          lift $ M.delete key validatorsMap
          modify $ Set.delete key
        else do
          -- update power in Validators map and ensure key is in key set
          lift $ M.insert key newPower validatorsMap
          modify $ Set.insert key
```

where `lift` is from `Control.Monad.Trans.Class`. It's equivalent to a `fold`, but allows for interleaving other effects as well. This means `lift` resolves to ` forall r a.  Sem r a -> StateT (Set PubKey_) (Sem r) a`

martyall:
Regarding new typing to avoid circular dependency, or what would usually be orphan instances , it’s a very common method. If you need to do it then go for it.

i think you should write up this issue for map with keys because it would be really useful. I remember it being a big annoyance in ethereum contracts. Regarding the grpc stuff, maybe you’ve seen but I have a branch to upgrade the grpc protobuf file from iavl. I just haven’t finished it yet...

UnitylChaos:
Okay. Did some stuff:

- Used StateT as recommended in EndBlock.
- Removed Keeper methods
- Switched ValidatorUpdate and PubKey to be newtypes around ABCI types which reuse existing proto serializations

Two things to note:
- I didn't switch KeySet to protobuf, because I expect this will end up being removed eventually anyway if #254 happens. (and didn't want to clutter things up with a new proto file for one temporary type, which isn't externally accessible)
- I also removed the Maybe from PubKey in the ABCI ValidatorUpdate definition. The actual proto has this as not nullable, and the Tendermint side expects it to not be null, so it seemed valid to expect the data type to always have a pubkey value.

hs-abci-sdk/src/Tendermint/SDK/Modules/Validators/Keeper.hs
```diff
@@ -0,0 +1,73 @@
...
+  => PubKey_
+  -> Sem r Word64
+getPowerOfF key =
+  fmap (fromMaybe 0) $ M.lookup key validatorsMap
+
+getQueuedUpdatesF
+  :: Members [ReadStore, Error AppError] r
+  => Sem r (Map.Map PubKey_ Word64)
+getQueuedUpdatesF = L.foldl (\m (ValidatorUpdate_ ValidatorUpdate{..}) ->
+  Map.alter (Just . fromMaybe (toWord validatorUpdatePower)) (PubKey_ validatorUpdatePubKey) m) Map.empty updatesList
```
blinky3713, line 60: What does it mean if two `ValidatorUpdate_`s have been queued for the same `PubKey_`? It seems like here we just take the one that happened last as the update and disregard whatever came before it?

martyall: What does it mean if two ValidatorUpdate_s have been queued for the same PubKey_? It seems like here we just take the one that happened last as the update and disregard whatever came before it?

hs-abci-sdk/src/Tendermint/SDK/Modules/Validators/Types.hs
```diff
@@ -0,0 +1,59 @@
...
+
+validatorsKeySetKey :: ByteString
+validatorsKeySetKey = "validatorsKeySet"
+
+
+newtype ValidatorUpdate_ = ValidatorUpdate_ ValidatorUpdate deriving (Eq, Generic)
+
+instance HasCodec ValidatorUpdate_ where
+  encode (ValidatorUpdate_ vu) = encodeMessage $ (vu ^. _Wrapped')
+  decode bs = bimap pack (ValidatorUpdate_ . (^. _Unwrapped')) $ decodeMessage bs
```
martyall: We don't really use things like `pack` because it leads to having to import like 10 different `String`-like modules throughout the project. We use the package [string-conversions](https://hackage.haskell.org/package/string-conversions) to abstract it away via the [cs method](https://hackage.haskell.org/package/string-conversions-0.4.0.1/docs/Data-String-Conversions.html#v:cs).

hs-abci-sdk/src/Tendermint/SDK/Modules/Validators/Types.hs
```diff
@@ -0,0 +1,59 @@
...
+
+newtype PubKey_ = PubKey_ PubKey deriving (Eq, Ord, Generic)
+
+instance RawKey PubKey_ where
+  rawKey = iso t f
+    where
+      t (PubKey_ p) = encodeMessage $ (p ^. _Wrapped')
+      f = PubKey_ . fromRight (PubKey "" "") . second (^. _Unwrapped') . decodeMessage
```
martyall, line 48: This is fine, but as a note I'm sort of remembering that there is actually no instance where the `from` method of the iso is being used, we're almost always using it as a value of type `:: Lens' a ByteString`. This is why the `PubKey "" ""` seems awkward, I will get around to fixing this soon.

martyall (changes requested):
OK looks good to me! Just a few small review comments, only one small change. Also need to Otherwise the majority of the work will be in figuring out #254 at some point. Great work!

Regarding weeder, I think you need to put the tope level `Tendermint.SDK.Modules.Validators` and generated protobuf modules in the `package.yaml`, similar to what you see [here](https://github.com/f-o-a-m/kepler/blob/master/hs-abci-sdk/package.yaml#L136-L150)

hs-abci-sdk/src/Tendermint/SDK/Modules/Validators/Types.hs
martyall: See #255

hs-abci-sdk/src/Tendermint/SDK/Modules/Validators/Keeper.hs
UnitylChaos: Yes that was the intended behavior here. I figured there were roughly 3 options of how to handle multiple updates to the same validator within a single block:
1. Don't allow it, which would require turning the updates array into a map to check if the pubkey already existed
2. Allow it, and apply only the first update
3. Allow it, and apply only the last update


Applying the last update made the most sense to me, as all the updates are replacements not relative changes. So the result of composing those replacements in order would be to just take the last one.

hs-abci-sdk/src/Tendermint/SDK/Modules/Validators/Types.hs
UnitylChaos: :+1: good to know, I too have been annoyed by keeping track of various string types, so having a way to handle that is nice.

hs-abci-sdk/src/Tendermint/SDK/Modules/Validators/Types.hs
UnitylChaos: re `PubKey "" ""`: 
I figured this would be okay, since the only time it would be deserializing would be if it was reading the bytes from the db. And they must have been serialized from the pubkey to get put in the DB, so it _should_ never fail. Definitely is awkward thought...

UnitylChaos:
So this is an interesting test failure:
```
Failures:

  test/Network/ABCI/Test/Types/MessagesSpec.hs:46:5: 
  1) Network.ABCI.Test.Types.Messages.Request.initChain Unwrapped -> Wrapped -> Unwrapped == identity
       Falsified (after 4 tests and 5 shrinks):
         ArbitraryMessage {unArbitraryMessage = {validators { }}}                                                

  To rerun use: --match "/Network.ABCI.Test.Types.Messages/Request/initChain/Unwrapped -> Wrapped -> Unwrapped == identity/"

  test/Network/ABCI/Test/Types/MessagesSpec.hs:46:5: 
  2) Network.ABCI.Test.Types.Messages.Response.initChain Unwrapped -> Wrapped -> Unwrapped == identity
       Falsified (after 2 tests and 2 shrinks):
         ArbitraryMessage {unArbitraryMessage = {validators { }}}

  To rerun use: --match "/Network.ABCI.Test.Types.Messages/Response/initChain/Unwrapped -> Wrapped -> Unwrapped == identity/"

  test/Network/ABCI/Test/Types/MessagesSpec.hs:46:5: 
  3) Network.ABCI.Test.Types.Messages.Response.endBlock Unwrapped -> Wrapped -> Unwrapped == identity
       Falsified (after 3 tests and 3 shrinks):
         ArbitraryMessage {unArbitraryMessage = {validator_updates { }}}

  To rerun use: --match "/Network.ABCI.Test.Types.Messages/Response/endBlock/Unwrapped -> Wrapped -> Unwrapped == identity/"

  test/Network/ABCI/Test/Types/MessagesSpec.hs:46:5: 
  4) Network.ABCI.Test.Types.Messages.FieldTypes.ValidatorUpdate Unwrapped -> Wrapped -> Unwrapped == identity
       Falsified (after 3 tests and 1 shrink):
         ArbitraryMessage {unArbitraryMessage = {}}
```

This is being caused by my changing the `ValidatorUpdate` type from: `Maybe PubKey` to `PubKey`. Interesting that it's only on the "Unwrapped -> Wrapped -> Unwrapped" side of the tests. 

My reading of this is that it can start with an "Unwrapped" protobuf value which doesn't have the key, and then when it tries to turn it into a ValidatorUpdate object it fails...

Is this basically why the value was `Maybe PubKey` in the first place? I sorta figured there was some way to cleanly tell it that a null PubKey could not be used to form a valid `ValidatorUpdate`, but I don't know how to tell the tests that.

I can also just switch the type back to `Maybe PubKey` if this is too much of a mess to fix.

martyall:
Yeah to be honest I’m not really sure, but try putting the Maybe back and if it passes we can get this in without worrying too much about it for now

hs-abci-sdk/src/Tendermint/SDK/Modules/Validators/Keeper.hs
```diff
@@ -57,17 +57,18 @@ getQueuedUpdatesF
   :: Members [ReadStore, Error AppError] r
   => Sem r (Map.Map PubKey_ Word64)
 getQueuedUpdatesF = L.foldl (\m (ValidatorUpdate_ ValidatorUpdate{..}) ->
-  Map.alter (Just . fromMaybe (toWord validatorUpdatePower)) (PubKey_ validatorUpdatePubKey) m) Map.empty updatesList
+  Map.alter (Just . fromMaybe (toWord validatorUpdatePower)) (toPK_ validatorUpdatePubKey) m) Map.empty updatesList
   where
     toWord (WrappedVal x) = fromIntegral x
+    toPK_ = PubKey_ . fromJust
```
martyall: Sorry, one last thing. We prefer not to use partial functions like this, its better to use things like `error` to at least print a message that can help debug

martyall: So like `fromMaybe (error “my error message”)`

UnitylChaos: Fair enough. This should never happen, because it's only unwrapping the Just that was put on by `queueUpdateF`, but since it's not represented in the types it's reasonable to be total / use an error message. Will do :+1:

martyall: Yeah it's just one of those things that would be nearly impossible to track down, especially since the assumption is so strong.

martyall (approved):

PR #256, opened 2021-01-12
Instances for Validators query support

UnitylChaos:
Bugfix for Validators module.

When trying to use the Validators module in an application, it complains about missing instances.
Specifically:
`No instance for (QueryData V.PubKey_)`
and
`No instance for (Tendermint.SDK.Codec.HasCodec
                         (Data.Map.Internal.Map V.PubKey_ GHC.Word.Word64))`

I gather this is a result of the query methods `GetPowerOf` (which uses PubKey_ as a search key), and `GetValidators` which returns the full `Map`.

### DIFF
--- a/hs-abci-sdk/src/Tendermint/SDK/Modules/Validators/Types.hs
+++ b/hs-abci-sdk/src/Tendermint/SDK/Modules/Validators/Types.hs
@@ -7,14 +7,16 @@ import           Data.Bifunctor                         (Bifunctor (bimap, secon
 import           Data.ByteString                        (ByteString)
 import           Data.ByteString.Lazy                   (toStrict)
 import           Data.Either                            (fromRight)
+import           Data.Map                               (Map)
 import           Data.ProtoLens                         (decodeMessage,
                                                          encodeMessage)
 import           Data.Set                               (Set)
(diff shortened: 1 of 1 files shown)


hs-abci-sdk/src/Tendermint/SDK/Modules/Validators/Types.hs
```diff
@@ -49,7 +51,19 @@ instance RawKey PubKey_ where
 
 
 instance A.ToJSON PubKey_
+instance A.ToJSONKey PubKey_
 instance A.FromJSON PubKey_
+instance A.FromJSONKey PubKey_
+
+instance HasCodec (Map PubKey_ Word64) where
+  encode = toStrict . A.encode
```
martyall: While we’re at it, let’s replace `toStrict` with `Data.String.Conversions.cs`

martyall (approved):
Just one little request that I’d prefer, otherwise looks good to me

hs-abci-sdk/src/Tendermint/SDK/Modules/Validators/Types.hs
UnitylChaos: :+1: good call.

PR #261, opened 2021-02-25
exposing length from array storage module

### LINKED ISSUE #258: Consider exporting length function in Array
Is there a reason this isn't exported ?

https://github.com/f-o-a-m/kepler/blob/a10aa3a84e33709289344c44e03a7cf022a1c2b5/hs-abci-sdk/src/Tendermint/SDK/BaseApp/Store/Array.hs#L232-L238

I can work around this by replicating with a Var in my module, or by using `length <$> Array.toList` but that seems rather inefficient.

martyall:
Fixes #258 

@jlandrews 

I also noticed that `foldl` was not exposed, so went ahead and did that.


UnitylChaos:
LGTM Thanks :+1:

PR #266, opened 2024-06-03
Fix height parameter in /block

alexfmpe:
Without this change I always get the latest block 

The go parameter is indeed [_heightPtr_](https://github.com/tendermint/tendermint/blob/v0.32.2/rpc/core/blocks.go#L230) but the query parameter is [_height_](https://github.com/tendermint/tendermint/blob/v0.32.2/rpc/core/blocks.go#L129). I'm not sure where and why the _Ptr_ bit gets removed, maybe because its type is a pointer?

The tests don't detect this because there's only one check and for the latest block
https://github.com/f-o-a-m/kepler/blob/bf75483fdc475eca3448eaf88489449167cdba21/hs-tendermint-client/kv-test/KVStore/Test/KVSpec.hs#L44-L47

Maybe there should be a few cases checking that the returned `blockMeta -> header -> height` matches the given height if below latest?

### DIFF
--- a/hs-tendermint-client/src/Network/Tendermint/Client.hs
+++ b/hs-tendermint-client/src/Network/Tendermint/Client.hs
@@ -110,13 +110,13 @@ block = RPC.remote (RPC.MethodName "block")
 
 -- https://github.com/tendermint/tendermint/blob/v0.32.2/rpc/core/blocks.go#L72
 data RequestBlock = RequestBlock
-  { requestBlockHeightPtr :: Maybe (FieldTypes.WrappedVal Int64)
+  { requestBlockHeight :: Maybe (FieldTypes.WrappedVal Int64)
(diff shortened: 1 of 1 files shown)


martyall (approved):
This repo is severely under-maintained so I'm excited that it's become relevant for you again. I would be happy to add you a maintainer if you're interested

martyall:
> Maybe there should be a few cases checking that the returned `blockMeta -> header -> height` matches the given height if below latest?

If you want to add that test, go for it. I am currently not in a state to be able to build and modify the project, so I support whatever you think is required here to make sure it's working for you're needs.

alexfmpe:
> I would be happy to add you a maintainer if you're interested

Yeah, I am - now and then I tweak this or that and it's been piling up in a fork but I rather not diverge

